PR #46, opened 2014-12-08
EXPERIMENTAL for hpux build.

kinichiro:
environment: HP-UX 11.31 ia64, gcc 4.7.1(HP AllianceOne version) or HP C/aC++

modified 3 files.

- configure.ac - add hpux section.
- crypto/Makefile.am - add hpux section.
- crypto/compat/arc4random.h - add hpux section.

added 1 file.

- crypto/compat/issetugid_hpux.c - using pstat_getproc() and get*id()
  hpux doesn't have issetugid(), and this is sample implementation.
  this should be verified.


busterb:
I'm doing the sync from CVS now to merge the rest of this.

busterb:
OK, merged up and we're green here as well. Thanks a lot, this will be in the next release.

kinichiro:
I could verify that both gcc and HP C/aC++ succeeded to build on hpux11i(11.31).
Thanks for your hard work.

Let me add 1 tip for how to build with HP C/aC++ compiler.

If you want to build with HP C/aC++ compiler instead of gcc,
you should modify CFLAGS in configure (it is about line 3018) like this.

    CFLAGS="-g +DD64 -D_XOPEN_SOURCE=600 -D__STRICT_ALIGNMENT"

busterb:
It looks like there are a couple of compiler warnings that show up for me as well:

This might caused some alignment issues:
```
gost/gostr341001_ameth.c: In function 'pub_decode_gost01':
gost/gostr341001_ameth.c:209: warning: dereferencing type-punned pointer will break strict-aliasi
ng rules
gost/gostr341001_ameth.c: In function 'priv_decode_gost01':
gost/gostr341001_ameth.c:419: warning: dereferencing type-punned pointer will break strict-aliasi
ng rules
```

I think it we should could just rename the defines in the openssl(1) code to avoid this.
```
apps.c:1387:1: warning: "BSIZE" redefined
In file included from /usr/include/limits.h:508,
                 from /usr/local/lib/gcc/ia64-hp-hpux11.31/4.2.3/include/limits.h:122,
                 from /usr/local/lib/gcc/ia64-hp-hpux11.31/4.2.3/include/syslimits.h:7,
                 from /usr/local/lib/gcc/ia64-hp-hpux11.31/4.2.3/include/limits.h:11,
                 from apps.c:135:
/usr/include/sys/param.h:234:1: warning: this is the location of the previous definition
ca.c:294:1: warning: "BSIZE" redefined
In file included from /usr/include/limits.h:508,
                 from /usr/local/lib/gcc/ia64-hp-hpux11.31/4.2.3/include/limits.h:122,
                 from /usr/local/lib/gcc/ia64-hp-hpux11.31/4.2.3/include/syslimits.h:7,
                 from /usr/local/lib/gcc/ia64-hp-hpux11.31/4.2.3/include/limits.h:11,
                 from ca.c:66:
/usr/include/sys/param.h:234:1: warning: this is the location of the previous definition
```

kinichiro:
Hi
I also checked compilation warnings with hp-gcc-4.7.1 on hpux11i(11.31).

I couldn't see 1st one (alignment issue).

    ...
    CC       gost/libcrypto_la-gostr341001.lo
    CC       gost/libcrypto_la-gostr341001_ameth.lo
    CC       gost/libcrypto_la-gostr341001_key.lo
    ...

This warning seems to occur with gcc-4.2.3 only,
and both line 209 and 419 of gostr341001_ameth.c are like this.

    X509_ALGOR_get0(NULL, &ptype, (void **)&pval, palg);

X509_ALGOR_get0() is defined in asn1/x_alogr.c, and `(void **)` should be removed.
Other xx_ameth.c (ex. dh/dh_ameth.c) does NOT use cast for `&pval`.


Then, I could see 2nd one (redefined BSIZE warning), yes.
And I think your resolution (rename BSIZE in apps.c and ca.c) is good.

Thanks.

PR #53, opened 2014-12-30
Turn on memory corruption hardening where available

jbarlow83:
Many compilers enable some kind of hardening by default, but this is not universally the case, and some more aggressive options are not always enabled.

There may be a performance impact to -fstack-protector-all by default, but Chromium uses this option when -fstack-protector-strong is not available, and I think that is appropriate for a critical security component.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,4 +8,4 @@ if ENABLE_LIBTLS
 pkgconfig_DATA += libtls.pc
 endif
 
-EXTRA_DIST = VERSION config
+EXTRA_DIST = VERSION config scripts
--- a/configure.ac
+++ b/configure.ac
@@ -80,7 +80,7 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([], [[
 AC_MSG_RESULT([$CLANG])
 AS_IF([test "x$CLANG" == "xyes"], [CLANG_FLAGS=-Qunused-arguments])
 
-# We want to check for compiler flag support, but there is no way to make
+# We want to check for compiler flag support. Prior to clang v5.1, there was no way to make
 # clang's "argument unused" warning fatal.  So we invoke the compiler through a
 # wrapper script that greps for this message.
 saved_CC="$CC"
--- a/configure.ac
+++ b/configure.ac
@@ -77,7 +77,7 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([], [[
 	[CLANG=yes],
 	[CLANG=no]
 )
-AC_MSG_RESULT([CLANG])
+AC_MSG_RESULT([$CLANG])
 AS_IF([test "x$CLANG" == "xyes"], [CLANG_FLAGS=-Qunused-arguments])
 
 # We want to check for compiler flag support, but there is no way to make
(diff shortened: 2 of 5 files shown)


busterb:
Thanks, I do like the results here!

A few more comments:

I think we need something in the scripts/wrap-compiler-for-flag-check file that explicitly says something like 'This file is in the public domain.' right at the top, based on the README here: https://github.com/kmcallister/autoharden

Have you looked at what libsndfile does? https://github.com/erikd/libsndfile/tree/master/Scripts

I'm not sure if some of the comments from @ToddMiller1 / @millert at http://mainisusuallyafunction.blogspot.com/2012/05/automatic-binary-hardening-with.html have been addressed. There don't seem to have been any commits to the repository since that blog post.

jbarlow83:
On Thu, Jan 1, 2015 at 10:33 AM, Brent Cook <notifications@github.com>
wrote:

> Thanks, I do like the results here!
>
> A few more comments:
>
> I think we need something in the scripts/wrap-compiler-for-flag-check file
> that explicitly says something like 'This file is in the public domain.'
> right at the top, based on the README here:
> https://github.com/kmcallister/autoharden
>
> Have you looked at what libsndfile does?
> https://github.com/erikd/libsndfile/tree/master/Scripts
>
Do you mean their "sanitize" script for clang? That looks like a good way
to make sanitize more available for testing.

> I'm not sure if some of the comments from @ToddMiller1
> <https://github.com/ToddMiller1> / @millert <https://github.com/millert>
> at
> http://mainisusuallyafunction.blogspot.com/2012/05/automatic-binary-hardening-with.html
> have been addressed. There don't seem to have been any commits to the
> repository since that blog post.
>
Re @ToddMiller1: His first comment was about getting -fPIE to work on
GCC/OS X 10.6. Since libressl is mainly a library I don't think -fPIE will
add much protection unless the openssl binary is statically linked.

Second comment is on improving the wrap-compiler-for-flag-check situation
using autoconf intrinsics.
However, it appears that the AC_LANG_WERROR([on/off/pop]) was proposed but
not implemented, so you have to manually implement the stack in a hacky way
that depends on internal variables in the AC_LANG_WERROR macro. It also
looks like OS X is not completely warning free so just setting
AC_LANG_WERROR in generally is not an option yet.

Another finding was that wrap-compiler-for-flag-check is mainly dealing
with a pre-clang 5.1 problem (about 1.5 years old). I think libsndfile does
not check for this case and probably gets the flags wrong on clang. I
thought the best thing to do here was document this fact so this workaround
can be reevaluated and removed when clang 5.1 is too old to worry about.

busterb:
Thanks for the explanation and for working on this.

busterb:
I made a couple of fixes changing the == compares to = to support systems running posix shells f2d68c7d21f11f3fa08ab821ccd4356fc863ccd4

I also found some issues building with GCC 4.2. Given the potential for other unexpected things to go haywire, I disabled the ERROR message for now in ec81c28219b3d3234ddf0af12d0579efa1a3d2a . We can re-enable that part later after the hardening flags have had a release for getting some field testing/feedback.

PR #58, opened 2015-01-10
Add code coverage flags to configure (./configure --enable-code-coverage)

jbarlow83:
To activate the feature, run
```
    ./configure --enable-code-coverage
```
and then `make` or possibly `make check`.

When enabled, optimization is switched off and the code is instrumented to generate data at both compile and runtime to measure coverage. A variety of open source programs are available for generating reports on coverage, using the data produced by `make check`. Obviously, the instrumented code runs slower and is not suitable for production.

If `make check-code-coverage` is run, this code will run `lcov` and `geninfo` to create HTML reports of code coverage. I have this feature working on OS X with Clang, but not Ubuntu 14.04 with GCC, the latter failing with an out of memory error (possibly a VM issue or older version of `lcov`).  

The `gcov` program is alternative to `lcov`; it generates plain text instead of HTML.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -197,3 +197,11 @@ man/*.1
 man/Makefile.am
 .gitmodules
 COPYING
+
+# Code coverage
+*.func.html
+*.gcov.html
+*.func-sort-c.html
+*-coverage/*
+*-coverage.info
+crypto/*-coverage.info.tmp
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,4 +8,6 @@ if ENABLE_LIBTLS
 pkgconfig_DATA += libtls.pc
 endif
 
+@CODE_COVERAGE_RULES@
+
 EXTRA_DIST = VERSION config scripts
--- a/apps/Makefile.am
+++ b/apps/Makefile.am
@@ -2,8 +2,9 @@ include $(top_srcdir)/Makefile.am.common
 
 bin_PROGRAMS = openssl
 
-openssl_CFLAGS = $(USER_CFLAGS)
-openssl_LDADD = $(PLATFORM_LDADD) $(PROG_LDADD)
+openssl_CFLAGS = $(USER_CFLAGS) $(CODE_COVERAGE_CFLAGS)
+openssl_LDFLAGS = $(CODE_COVERAGE_LDFLAGS)
+openssl_LDADD = $(PLATFORM_LDADD) $(PROG_LDADD) 
 openssl_LDADD += $(top_builddir)/ssl/libssl.la
 openssl_LDADD += $(top_builddir)/crypto/libcrypto.la
 
--- a/configure.ac
+++ b/configure.ac
@@ -272,7 +272,12 @@ AM_COND_IF([ENABLE_LIBTLS], [AC_CONFIG_FILES([libtls.pc])])
 LT_INIT
 
 # gcov coverage reporting
-AX_CODE_COVERAGE
+AX_CODE_COVERAGE([$CLANG])
+AS_IF([ test "x$enable_code_coverage" = "xyes" ], [
+	changequote({,})
+	USER_CFLAGS=`echo "$USER_CFLAGS" | $SED -e 's/-O[0-9]*//g'`
+	changequote([,])
+	])
 
 AC_CONFIG_FILES([
 	Makefile
--- a/configure.ac
+++ b/configure.ac
@@ -271,6 +271,9 @@ AM_COND_IF([ENABLE_LIBTLS], [AC_CONFIG_FILES([libtls.pc])])
 
 LT_INIT
 
+# gcov coverage reporting
+AX_CODE_COVERAGE
+
 AC_CONFIG_FILES([
 	Makefile
 	include/Makefile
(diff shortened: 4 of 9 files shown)


kinichiro:
Hi.
It is very cool.
I could `./configure --enable-code-coverage` and `make check-code-coverage` 
successfully on this environment.
- CentOS6 (2.6.32-504.1.3.el6.x86_64)
- gcc (GCC) 4.4.7
- LCOV version 1.11

Does this coverage data connect with https://coveralls.io/ ?
Thanks.

busterb:
I guess I need to try this to see how it works. It's supposed to show how much code is exercised in unit tests?

jbarlow83:
@kinichiro It would be possible to connect with coveralls using https://github.com/eddyxu/cpp-coveralls. Although it looks like the entry level service costs money.

@busterb Yes. `make check-code-coverage` will generate HTML reports and display a file:/// URL for local viewing after it runs all of the `make check` unit tests. The reports show how many times an individual code line was hit, percentage of lines hit in a file, etc. (You might have to `make clean` before rerunning configure.) If you're wondering, the current coverage results are not great.

jbarlow83:
(Clicked close by mistake...)

jbarlow83:
image

busterb:
cool - sounds like a good incentive to add some more tests

busterb:
I think you have invoked some non-portable make behavior here. This breaks the build on OpenBSD for instance, and causes some automake warnings:

```
tests/Makefile.am:11: warning: 'CFLAGS' is a user variable, you should not override it;
tests/Makefile.am:11: use 'AM_CFLAGS' instead
tests/Makefile.am:12: warning: 'LDFLAGS' is a user variable, you should not override it;
tests/Makefile.am:12: use 'AM_LDFLAGS' instead
 /bin/sh ./config.status
config.status: creating Makefile
config.status: creating include/Makefile
config.status: creating include/openssl/Makefile
config.status: creating crypto/Makefile
config.status: creating ssl/Makefile
config.status: creating tls/Makefile
config.status: creating tests/Makefile
config.status: creating apps/Makefile
config.status: creating man/Makefile
config.status: creating libcrypto.pc
config.status: creating libssl.pc
config.status: creating openssl.pc
config.status: executing depfiles commands
config.status: executing libtool commands
Making all in ssl
Making all in tls
Making all in include
Making all in openssl
Making all in apps
Making all in tests
Making all in man
*** Parse error in /home/bcook/projects/libressl/portable: Missing dependency operator (Makefile:913)
*** Parse error: Need an operator in 'else' (Makefile:916)
*** Parse error: Need an operator in 'endif' (Makefile:918)
*** Parse error: Missing dependency operator (Makefile:922)
*** Parse error: Need an operator in 'else' (Makefile:928)
*** Parse error: Need an operator in 'endif' (Makefile:930)
*** Parse error: Missing dependency operator (Makefile:935)
*** Parse error: Need an operator in 'endif' (Makefile:941)
*** Error 1 in /home/bcook/projects/libressl/portable (Makefile:474 'all-recursive')
```

jbarlow83:
Does gmake work on OpenBSD? It seems that the "missing dependency operator"
has to do with a BSD make/GNU make difference.

I tested OS X and Ubuntu.

busterb:
It's more portable to just shell out like so:

```
--- a/m4/ax_code_coverage.m4
+++ b/m4/ax_code_coverage.m4
@@ -193,35 +193,35 @@ code_coverage_quiet_0 = --quiet
 
 # Use recursive makes in order to ignore errors during check
 check-code-coverage:
-ifeq ($(CODE_COVERAGE_ENABLED),yes)
-       -$(MAKE) $(AM_MAKEFLAGS) -k check
-       $(MAKE) $(AM_MAKEFLAGS) code-coverage-capture
-else
-       @echo "Need to reconfigure with --enable-code-coverage"
-endif
+       if [[ "$(CODE_COVERAGE_ENABLED)" = yes ]]; then \
+               -$(MAKE) $(AM_MAKEFLAGS) -k check; \
+               $(MAKE) $(AM_MAKEFLAGS) code-coverage-capture; \
+       else \
+               @echo "Need to reconfigure with --enable-code-coverage"; \
+       fi
```

After fixing the rest of the issues, I started to wonder who other than a developer would really need this  anyway (that is, do we really need to ship this as part of the tarball release?). And if that's so, I wonder if this mechanism could just be implemented as a simple shell script like dist.sh. Think about it for a bit. If not, I can submit a PR back for your branch on Sunday that includes the fixes - I'm going to be away-from-keyboard most of tomorrow.

busterb:
Hi!

Rather than pulling in a few hundred lines into the build system that doesn't work well on non-Linux systems, I just wrote a short script to basically do the same thing: 3cb34ee99f9be3e866ca63110fec90bc104fe800. An advantage here is it even can be used retroactively on previous releases.

It doesn't have fancy autodetection and clang support, but it works well enough as a dev tool. Thanks for doing the work here, but I think I will pass on integrating this into the build directly.

twphoenix:
Hey!

 

Important message, please visit <http://somaticyoga.com/break.php?gj>

 

rene.l@fictionshare.net

PR #60, opened 2015-01-31
allow compiling on Linux with non-GNU libc

gandaro:
At least one alternative libc,  musl, provides `getauxval`
as well.  But musl does not provide `gnu_get_libc_version`
or `<gnu/libc-version.h>`.  This patch keeps the check for
older versions of glibc, but allows compiling with musl,
too.

### DIFF
--- a/crypto/compat/issetugid_linux.c
+++ b/crypto/compat/issetugid_linux.c
@@ -4,10 +4,11 @@
  */
 
 #include <errno.h>
-#include <gnu/libc-version.h>
+
+#ifdef __GLIBC__
 #include <string.h>
-#include <sys/types.h>
-#include <unistd.h>
+#include <gnu/libc-version.h>
+#endif
 
 /*
  * Linux-specific glibc 2.16+ interface for determining if a process was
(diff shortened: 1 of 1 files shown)


busterb:
I'm pretty sure this is innecessary - musl has issetugid since 1.1.4, so this function should be unused entirely. What version are you building against?

busterb:
I tested with the latest uclibc and musl, and they should work fine without any changes. Using the getauxval interface makes it difficult to emulate issetugid on any libc other than glibc without false-positives, so I recommend using the native versions rather than attempting an emulation. Thanks!

gandaro:
You are correct.  I don’t know why compiling did not work for me before.

busterb:
Glad to hear it!

Just a guession, but maybe your autom4te.cache was stale from a previous build with glibc? I've gotten weird behavior like that when I forgot to 'make distclean' when switching between toolchains when doing in-tree builds. Out-of-tree builds can be useful for this, e.g. untar the source, mkdir build, cd build, ../libressl-blah/configure

PR #61, opened 2015-03-03
Add preliminary Cygwin support

github-cygwin:
Hi,

I'd like to add support for Cygwin to libressl portable, slowly and
carefully.  For a start I'd like to propose the below patch.  It allows
to build with the default hardening without having to specify
--enable-windows-ssp.  Other than that change, libressl-portable builds
out of the box on Cygwin 1.7.35, the upcoming release later this week. 
The reason it needs 1.7.35 is the existence of the issetugid function.
The functionality existed for a long time, it was just never exported so far.

Unfortunately the testsuite doesn't build on Cygwin, because the testsuite
requires SA_ONSTACK and sigaltstack to work, functionality not available
yet on Cygwin.  But that's another issue.

Thanks,
Corinna

Signed-off-by: Corinna Vinschen <github@cygwin.de>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,6 +13,9 @@ AC_SUBST([USER_CFLAGS], "-O2 $CFLAGS")
 CFLAGS="$CFLAGS -Wall -std=gnu99 -g"
 
 case $host_os in
+	*cygwin*)
+		HOST_OS=cygwin
+		;;
 	*darwin*)
 		HOST_OS=darwin
 		HOST_ABI=macosx
@@ -53,6 +56,7 @@ case $host_os in
 	*) ;;
 esac
 
+AM_CONDITIONAL([HOST_CYGWIN],  [test x$HOST_OS = xcygwin])
 AM_CONDITIONAL([HOST_DARWIN],  [test x$HOST_OS = xdarwin])
 AM_CONDITIONAL([HOST_FREEBSD], [test x$HOST_OS = xfreebsd])
 AM_CONDITIONAL([HOST_HPUX],    [test x$HOST_OS = xhpux])
@@ -163,7 +167,7 @@ AS_IF([test "x$enable_hardening" = "xyes"], [
 				AC_MSG_WARN([compiler does not appear to support stack protection])
 			)
 		)
-		AS_IF([test "x$HOST_OS" = "xwin"], [
+		AS_IF([test "x$HOST_OS" = "xwin" -o "x$HOST_OS" = "xcygwin"], [
 			AC_SEARCH_LIBS([__stack_chk_guard],[ssp])
 		])
 	])
--- a/configure.ac
+++ b/configure.ac
@@ -167,7 +167,7 @@ AS_IF([test "x$enable_hardening" = "xyes"], [
 				AC_MSG_WARN([compiler does not appear to support stack protection])
 			)
 		)
-		AS_IF([test "x$HOST_OS" = "xwin" -o "x$HOST_OS" = "xcygwin"], [
+		AS_IF([test "x$HOST_OS" = "xwin"], [
 			AC_SEARCH_LIBS([__stack_chk_guard],[ssp])
 		])
 	])
--- a/update.sh
+++ b/update.sh
@@ -198,9 +198,6 @@ for i in `find $libcrypto_regress -name '*.c'`; do
 	 $CP "$i" tests
 done
 
-# the BIO tests rely on resolver results that are OS and environment-specific
-rm tests/biotest.c
-
 # copy libc tests
 $CP $libc_regress/arc4random-fork/arc4random-fork.c tests/arc4randomforktest.c
 $CP $libc_regress/explicit_bzero/explicit_bzero.c tests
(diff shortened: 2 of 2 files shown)


github-cygwin:
It appears that adding -lssp is unnecessary.  The problem was that I autreconf'ed the project
on Fedora 21, which still comes with an old libtool version 2.4.2.  Using the latest libtool
version 2.4.6, the -lssp is added automatically to the command line, as is correct and desired.

So this part of the patch is not required.

How shall I go forward, retracting this pull request and generate a new one with just
the HOST_OS addition?

Thanks in advance,
Corinna

bcook-r7:
You can push the fix on this branch and the PR will update. Squashing the commits and doing a force push works fine too.

github-cygwin:
I pushed the change.  Thanks!

May I take the opportunity to ask another question?  I'm having the patches required to run
the testsuite successfully on Cygwin.  The patch only changes tests/Makefile.am, but that
file is part of the OpenBSD part.  How can I contribute the patch?  The comment on
https://github.com/libressl-portable/openbsd indicates that a pull request there
won't help...

Thanks,
Corinna

bcook-r7:
Sure, you can post it there and refer to it here. The OpenBSD tree is currently in lock/stabilization mode for the next release, but  it should open up in a few weeks and I can help shepherd it in from the PR.

github-cygwin:
Ouch, hang on.  I just realized that tests/Makefile.am is actually generated in update.sh.
So it kind of is part of portable., but a patch will be a bit tricky...

bcook-r7:
I'd be fine if we moved the explicit_bzero stuff to Makefile.am.tpl if it makes it easier, since its already handled strangely in update.sh as well.

github-cygwin:
That would be nice.  For Cygwin, explicit_bzero and pidwraptest have to be dropped, the latter one
only stopping if the PID is getting a value impossible on Cygwin (the grandparent PID is blocked
inside Cygwin in this scenario).
Do you want me to provide a patch?

Thanks,
Corinna

bcook-r7:
It might be good enough to have it off by default again, like biotest. It causes pain on other platforms as well as a regular test.

github-cygwin:
Hi Brent,

I reworked update.sh a bit to use the tests_disabled array which was already tested in the second
loop creating the Makefile.am content.  I rearranged the loops a bit so the tests_disabled content
is checked first and no ouput is generated if the test is in that array.   I moved the biotest test into
tests_disabled as well, so this is now done in a unified, extensible way.

Hope that's ok.

Thanks,
Corinna

busterb:
I think we can make the explicit bzero inclusion dependent on SA_ONSTACK existing automatically with an AC_COMPILE_IFELSE check. Would you be able to make that work?

github-cygwin:
Hi Brent,

On Mar  6 08:59, Brent Cook wrote:
> I think we can make the explicit bzero inclusion dependent on
> SA_ONSTACK existing automatically with an AC_COMPILE_IFELSE check.
> Would you be able to make that work?

Hmm, maybe, but this may take a while since I have other stuff to work
on the next couple of days.

That doesn't affect the pull reqeust to introduce the cygwin target,
right?


Corinna

busterb:
no problem, I can take care of it.

busterb:
I pulled your commits, and made a few changes after to continue testing explicit_bzero where the signal stack is available. The other two flaky tests are guarded by --enable-extratests.

It looks like there still may be issues enabling stack guard without explicitly linking the library, at least on the version of cygwin I was using, and issetugid looks like it will need a cygwin-specific implementation as well.

Thank you!

github-cygwin:
Hi Brent,

On Mar 22 05:59, Brent Cook wrote:
> I pulled your commits, and made a few changes after to continue
> testing explicit_bzero where the signal stack is available. The other
> two flaky tests are guarded by --enable-extratests.

Thanks!

> It looks like there still may be issues enabling stack guard without
> explicitly linking the library, at least on the version of cygwin I
> was using,

Do you mean -fstack-protector and linking against -lssp?  If so, then
this is not a Cygwin problem, but rather a problem of using a too old
libtool which strips the -fstack-protector-FOO.

I stumbled over this myself when I provided my first patch, but a
collegue of mine pointed out that you have to use at least libtool
version 2.4.2 (I think).  Current upstream is 2.4.6 and that's what's
provided by the Cygwin distro as well.

If you created the libtool files with an older version you should
re-libtoolize -f with a more recent libtool version which doesn't
strip the -fstack-protector-FOO options from the command line and
then GCC will add -lssp automatically, as desired.

> and issetugid looks like it will need a cygwin-specific
> implementation as well.

You're using a non-recent Cygwin, apparently.  I added the issetugid
entry point to the current release version 1.7.35.  As I wrote, the
functionality already existed for a long time, it was just never
exported but rather used internally only.


Thanks,
Corinna

github-cygwin:
On Mar 23 09:44, github-cygwin wrote:
> Hi Brent,
> 
> On Mar 22 05:59, Brent Cook wrote:
> > I pulled your commits, and made a few changes after to continue
> > testing explicit_bzero where the signal stack is available. The other
> > two flaky tests are guarded by --enable-extratests.
> 
> Thanks!
> 
> > It looks like there still may be issues enabling stack guard without
> > explicitly linking the library, at least on the version of cygwin I
> > was using,
> 
> Do you mean -fstack-protector and linking against -lssp?  If so, then
> this is not a Cygwin problem, but rather a problem of using a too old
> libtool which strips the -fstack-protector-FOO.
> 
> I stumbled over this myself when I provided my first patch, but a
> collegue of mine pointed out that you have to use at least libtool
> version 2.4.2 (I think).  Current upstream is 2.4.6 and that's what's
> provided by the Cygwin distro as well.

No, sorry.  libtool 2.4.2 is not sufficient.  Some in-between version
after 2.4.2 did it anyway.  2.4.6 definitely does the right thing.


Corinna

busterb:
Updated and it works great now. Cheers.

PR #67, opened 2015-03-08
Reuse _OPENBSD_SOURCE namespace on NetBSD (>=8.x)

krytarowski:
It enables (for now) the system-wide reallocarray(3) and strtonum(3).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -34,6 +34,7 @@ case $host_os in
 		;;
 	*netbsd*)
 		HOST_OS=netbsd
+		CFLAGS="$CFLAGS -D_OPENBSD_SOURCE"
 		;;
 	*openbsd*)
 		HOST_ABI=elf


busterb:
Is NetBSD's reallocarray actually the same? It appears to be a wrapper for reallocarr now.

krytarowski:
They ought to be the same. If it behaves differently please let me know!

tedu:
The NetBSD implementation of reallocarray is dangerously broken. If either size is 0, it will free the existing pointer and return NULL. Then the caller will free the pointer again.

jonathangray:
The NetBSD implementation of strtonum isn't implemented with strtoll either but with strtoi with base 0 which accepts leading 0x for hex and leading 0 for octal.  These are different semantics to what is expected.  LibreSSL Portable should stick with the compat versions for systems that don't implement functions correctly.

krytarowski:
@tedu @jonathangray thank you! I will double check it propose a fix in NetBSD for these two functions.

krytarowski:
Should be fixed.

busterb:
To verify, what version of NetBSD should I be testing this with? Not 6.1.5 I presume.

krytarowski:
The ``_OPENBSD_SOURCE`` namespace is under development in current, it will be released as NetBSD-8.0.

We can hold a bit with the tests, I'm going to push to NetBSD additional compat functions: ``timingsafe_bcmp(3)`` and ``explicit_bzero(3)``.

The current uses of ``timingsafe_memcmp(3)`` could be replaced with ``timingsafe_bcmp(3)`` without any issue. There is no need for retaining ``timingsafe_memcmp(3)`` in LibreSSL as everywhere we check for ``== 0`` or ``!= 0``.

PR #71, opened 2015-03-20
support linux uClibc builds

mostynb:
Please consider this small patch, which adds support for linux uClibc builds.

### DIFF
--- a/crypto/compat/issetugid_linux.c
+++ b/crypto/compat/issetugid_linux.c
@@ -4,7 +4,9 @@
  */
 
 #include <errno.h>
+#if defined(__GLIBC__) && !defined(__UCLIBC__)
 #include <gnu/libc-version.h>
+#endif
 #include <string.h>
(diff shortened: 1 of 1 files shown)


busterb:
Hi. issetugid was integrated into uclibc last year. See #60

busterb:
We should add a list of compatible C library versions to the docs to help here.

mostynb:
(I have adjusted the title to be more precise...)

The point of this file is to add support for linux libc's (like glibc) that don't have issetugid.  Without this patch, it only works for glibc 2.16+, with this patch it also works for uClibc versions before issetugid was added.  IMO this is useful because it often takes quite a long time for embedded linux devices to be upgraded to newer libc's (most often when the device is replaced by a newer model).

I can help put together a list of supported (linux) libc versions with and/or without this patch if you like- is the top-level README file the preferred location for this?

PR #75, opened 2015-03-27
Don't copy the source files if they haven't changed

carlosmn:
If the source files in the OpenBSD version have not changed since the
last time update.sh was run, the script will still override them,
updating the mtime and making the build system think it needs to rebuild
the whole library even if a single file changed.

Change the copy command to compare the files and only overwrite the
target if it has changed.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -52,7 +52,27 @@ do_mv() {
 		rm -f "$1"
 	fi
 }
-CP='cp -p'
+
+do_cp() {
+	if [ $# -lt 2 ]
+	then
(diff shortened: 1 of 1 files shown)


carlosmn:
It turns out that the `cp -p` does work correctly with the current script, so I'm not quite sure what I was doing to need a rebuild of everything. This may just add coplexity for no reason.

busterb:
Perhaps it was a timezone issue causing the source timestamps to be temporarily ahead of your system clock.

busterb:
Yeah, let's table this. Thanks.

PR #82, opened 2015-04-27
modify for HP-UX build, choose correct CFLAGS for gcc.

kinichiro:
this PR is for https://github.com/libressl-portable/portable/issues/81
Thanks.


busterb:
Thanks @kinichiro

PR #84, opened 2015-05-07
add app tests from regress/usr.bin/openssl

busterb:
These are added directly rather than imported by update.sh since they require local modifications and its not worth breaking everyone's git forks yet to import them through cvs2git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ tests/pbkdf2*
 tests/*.pem
 tests/testssl
 tests/*.txt
+!tests/optionstest.c
 
 # ctags stuff
 TAGS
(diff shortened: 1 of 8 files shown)


kinichiro:
I confirmed that these additional regression tests works fine with
CentOS7, HP-UX 11.31(GCC and CC) and Solaris 11.2(i86pc).
<pre>
============================================================================
Testsuite summary for libressl 2.2.0
============================================================================
# TOTAL: 50
# PASS:  50
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
</pre>

busterb:
Thanks for your help @kinichiro :)

PR #89, opened 2015-05-14
adding support for bitrig

dhuseby:


busterb:
I primarily added openbsd to -portable so I could build the distribution tarball on openbsd, but thanks for the patch.

PR #104, opened 2015-07-06
patch headers to avoid redefinitions on windows

scott-parker:


busterb:
Thanks. This does cause issues using the Windows crypto APIs, doesn't it (though I've seen similar defines in various apps). I wonder if we should inject a warning of some sort into the compiler output, just so it doesn't cause any surprises.

scott-parker:
It was causing issues when compiling with visual studio. If anything includes wincrypt.h there will be errors due to thing such as X509_NAME being redefined. But a compiler warning of some sort would be useful.

busterb:
I committed a modified version that generates a warning, and only does the substitution if wincrypt.h was included (to avoid hiding unintentional conflicts with regular defines). Thanks!

PR #116, opened 2015-08-04
disable strict aliasing on HP-UX C/aC++ compiler

kinichiro:
to disable strict aliasing on HP-UX C/aC++, `+Otype_safety=off` is right.
`+Otype_safety=strong` forces ANSI aliasing.


busterb:
Thanks for the fix. The different levels of aliasing support in the HP-UX C/aC++ compiler do present a confusing choice compared with others.

kinichiro:
Thanks, this fix works fine.
And, yes, it is a bit confusing.

PR #124, opened 2015-08-18
Add install targets and shared libraries to CMake

TheNietsnie:
See https://github.com/libressl-portable/portable/issues/123


bob-beck:
Brent in principle this is nice for linux although we'd have to parameterize the versions I expect.. As long as it doesn't break the windows build.

busterb:
Thanks! I made a change 61fe482d467cd79870275905f456f3cf7705918f to pull the version info from our existent VERSION files - this automates things a bit more.

PR #165, opened 2015-12-11
Replace STDIN_FILENO with _fileno

anthonynovatsis:
Replace STDIN_FILENO with _fileno as STDIN_FILENO results in compile
errors with Visual Studio 2015 (using CMake).


busterb:
Thanks. I originally didn't think that GetStdHandle would work when stdin is a pipe, but it seems like that works fine. I think it's better just to revert back to that but to keep the GetFileType check. Will fix shortly.

PR #183, opened 2016-04-04
fix cmake on HP-UX

kinichiro:
- CMakeLists.txt
  * add OS specific compiler flags and library
  * add checking size of time_t
  * add checking memmem()

- tests/CMakeLists.txt
  * add if(HAVE_MEMMEM) for explicit_bzero
  * add checking SMALL_TIME_T for rfc5280time

- crypto/CMakeLists.txt
  * add getentropy_hpux.c

- tls/CMakeLists.txt
  * fix checking strsep


kinichiro:
I confirmed this fix on HP-UX 11.31 (ia64) with cmake 3.0.2.
Both HP C/aC++ A.06.25 and gcc 4.7.1 worked fine.

busterb:
Looks good, thanks.

PR #184, opened 2016-04-05
execute tests that require srcdir environment variable with cmake

kinichiro:
- uncomment procedures for aeadtest, evptest, pq_test, ssltest, testdsa and testrsa
- add set_tests_properties() for setting environment variable srcdir
- tweak openssl path in ssltest.sh, testdsa.sh, testenc.sh and testrsa.sh


busterb:
Thanks!

Issue #178, opened 2016-03-13
AES-NI does not get enabled if using cmake, Fedora 22 Linux x86_64


Safari77:
But if building using ```./configure``` , AES-NI is enabled.
Used ```openssl speed -evp aes128``` and ```perf top``` to check.

busterb:
Correct, none of that is wired in for cmake. Care to submit a patch?

busterb:
This is fixed now.

PR #188, opened 2016-04-08
modify cmake to build ASM

kinichiro:
- add functionality compiling ASM with cmake
- to enable ASM, `cmake -DENABLE_ASM=on ..`


kinichiro:
relates to issue #178

PR #189, opened 2016-04-11
add cmake build options

kinichiro:
- add cmake build options as configure provides
  * -DENABLE_ASM (default ON)
  * -DENABLE_EXTRATESTS (default OFF)
  * -DENABLE_NC (default OFF)
  * -DOPENSSLDIR (default ${CMAKE_INSTALL_PREFIX}/etc/ssl)

- add biotest and pidwraptest if ENABLE_EXTRATESTS is ON

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,17 @@ string(STRIP ${TLS_VERSION} TLS_VERSION)
 string(REPLACE ":" "." TLS_VERSION ${TLS_VERSION})
 string(REGEX REPLACE "\\..*" "" TLS_MAJOR_VERSION ${TLS_VERSION})
 
+option(ENABLE_ASM "Enable assembly" ON)
+option(ENABLE_EXTRATESTS "Enable extra tests that may be unreliable on some platforms" OFF)
+option(ENABLE_NC "Enable installing TLS-enabled nc(1)" OFF)
+set(OPENSSLDIR ${OPENSSLDIR} CACHE PATH "Set the default openssl directory" FORCE)
+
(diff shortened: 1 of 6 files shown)


kinichiro:
cmake build error occurs on macosx both clang and gcc, configure seems ok.
<pre>
...
[ 94%] Built target aeadtest
Scanning dependencies of target aes_wrap
[ 94%] Building C object tests/CMakeFiles/aes_wrap.dir/aes_wrap.c.o
Linking C executable aes_wrap
Undefined symbols for architecture x86_64:
  "_OPENSSL_ia32cap_P", referenced from:
      _AES_cbc_encrypt in libcrypto.a(aes-macosx-x86_64.s.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [tests/aes_wrap] Error 1
make[1]: *** [tests/CMakeFiles/aes_wrap.dir/all] Error 2
make: *** [all] Error 2
...
</pre>

kinichiro:
found the resolution for "Undefined symbols".
variable OPENSSL_ia32cap_P in crypto/cryptlib.c is not initialized.
http://stackoverflow.com/questions/19398742/os-x-linker-unable-to-find-symbols-from-a-c-file-which-only-contains-variables

I added compiler flag `-fno-common` if CMAKE_SYSTEM_NAME is "Darwin".
This link error is NOT occured on OSX if crypto library is dynamic.
That is why configure build succeeds on OSX.
On Linux (CentOS7 x86_64), this link error is NOT occured even if crypto library is static.

PR #192, opened 2016-04-21
fix ld warning "attempted multiple inclusion of file" on Solaris

kinichiro:
- To avoid ld warning on Solaris, use abs_top_builddir in Makefile.am

On Solaris11.3(i386) I'm getting ld warning "attempted multiple inclusion of file" like below.
<pre>
...
  CCLD   openssl
ld: warning: file ../../crypto/.libs/libcrypto.so: linked to /work/github/libressl-2.4.0/crypto/.libs/libcrypto.so: attempted multiple inclusion of file
...
</pre>
This warning comes from Solaris ld can't tell `../../crypto/.libs/libcrypto.so` and `/work/github/libressl-2.4.0/crypto/.libs/libcrypto.so` are the same library.

openssl is defined that it should be linked with $(top_builddir)/ssl/libssl.la and $(top_builddir)/crypto/libcrypto.la in apps/openssl/Makefile.am .
And libtool seems try to link libraries at dependency_libs in .la files, too.

In ssl/libssl.la, libcrypto.la is described as absolute path at dependency_libs, like this.
<pre>
...
# Libraries that this one depends upon.
dependency_libs=' /work/github/libressl-2.4.0/crypto/libcrypto.la -lnsl -lsocket -lresolv'
...
</pre>
Then absolute path libcrypto and relative path libcrypto bring "attempted multiple inclusion of file".

To avoid this ld warnings, I would like to suggest using `abs_top_builddir` instead of `top_builddir` in Makefile.am.

Regards,

### DIFF
--- a/apps/nc/Makefile.am
+++ b/apps/nc/Makefile.am
@@ -12,9 +12,9 @@ EXTRA_DIST = nc.1
 EXTRA_DIST += CMakeLists.txt
 
 nc_LDADD = $(PLATFORM_LDADD) $(PROG_LDADD)
-nc_LDADD += $(top_builddir)/crypto/libcrypto.la
-nc_LDADD += $(top_builddir)/ssl/libssl.la
-nc_LDADD += $(top_builddir)/tls/libtls.la
+nc_LDADD += $(abs_top_builddir)/crypto/libcrypto.la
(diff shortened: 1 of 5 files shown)


busterb:
Nice trick, I didn't know that existed!

PR #197, opened 2016-05-31
pc: add platform-specific libs to Libs.private

wiiaboo:
Fixes compilations including libressl static libraries in MinGW.

Signed-off-by: Ricardo Constantino (:RiCON) <wiiaboo@gmail.com>


busterb:
Good catch, thank you.

Issue #146, opened 2015-10-02
Proposal: add tls_connect_fns and tls_accept_fns for 


krono:
Currently, `libtls` supports establishing connections
 * directly (`tls_connect`/`tls_connect_servername`)
 * with given read/write-fds (`tls_connect_fds`/`tls_accept_fds`)
 * with a given socket (`tls_connect_socket`/`tls_accept_socket`) (which seems to be a shim for the fd ones)

I propose to add a new way: `tls_connect_fns` and `tls_accept_fns` that act similar to `SSL_set_bio` and provide a callback-based interface.

## Rationale

In some circumstances, the socket or file descriptor for the connection might not be directly available. One such scenario can be a custom networking implementation in a programming language.
This happens, for example, in Squeak/Smalltalk. A dedicated plugin is responsible for networking and _another_ plugin (currently using the openssl api) is responsible for encryption. However, the networking plugin allows a buffer-based access to the socket (without ever exposing their I/O fds or socket fds).
Hence, currently, a BIO-Based approach is used
(for example, in https://github.com/itsmeront/squeakssl/blob/master/src/unix/sqUnixOpenSSL.c , lines
 * 208 (connection, writing to the socket)
 * 291 (accepting a connection, writing to the socket)
 * 80 (setting the bio for a SSL context)
 Note how the functions are called with `src` and `dst` buffers that the SSL implementation has to interact with)

PS: If I should use other means of communication for this proposal, please direct me there.

btrask:
I discussed this with Doug Hogan over email. Here was my proposal:

>     typedef ssize_t (*tls_read_cb)(void *_ctx, void *_buf, size_t _buflen);
>     typedef ssize_t (*tls_write_cb)(void *_ctx, const void *_buf,
>         size_t _buflen);
>     
>     int tls_accept_cbs(struct tls *_ctx, struct tls **_cctx,
>         tls_read_cb _cb_read, tls_write_cb _cb_write);
>     int tls_connect_cbs(struct tls *_ctx, tls_read_cb _cb_read,
>         tls_write_cb _cb_write, const char *_servername);
> 
> [...] The read and write cbs would work basically like `tls_read` and `tls_write` (but implemented by the user), and could either block or return poll in/out (only one each).

He sounded receptive and said they were doing to consider it. I offered to send a message to the mailing list but he said he wanted to keep bikeshedding to a minimum.

[...]

PR #203, opened 2016-06-28
Portable code part for callbacks in accept/connect

krono:
see https://github.com/libressl-portable/openbsd/pull/64
as propsal for #146


krono:
\o/
Thanks :)

PR #204, opened 2016-07-05
include OCSP test script

kinichiro:
to include ocsptest.sh into distribution tar.gz


busterb:
thanks, the distribution usually enables extra tests, so I didn't hit this in testing.

PR #212, opened 2016-07-19
Fix typo in USE_BUILTIN_ARC4RANDOM check

celan69:
Solaris 11 recently introduced a builtin arc4random in libc which fails the tests in "make check". Found USE_BUILTIN_ARC4RANDOM, but could not get it to work. Apparently, there is a typo in the configure logic rendering USE_BUILTIN_ARC4RANDOM ineffective.


busterb:
Of course, thanks for the update. I'll need to update and see what the issue is in the latest solaris 11.

celan69:
Note that by "recently" I am referring to one of the latest patch updates (e.g. 11.3.4.9) to Solaris 11.3 - the stock 11.3 does not yet have an arc4random in libc and everything was fine.

PR #213, opened 2016-08-02
Set _PATH_SSL_CA_FILE for the OCSP tests in the cmake build system

geoffbeier:
Currently on cmake builds, make test shows the OCSP tests as failing if /etc/ssl/cert.pem does not exist. This change to CMakeLists.txt defines _PATH_SSL_CA_FILE to a sane value if either OPENSSLDIR or CMAKE_INSTALL_PREFIX is set so that these tests can run successfully without a system-installed libressl.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -196,6 +196,11 @@ add_test(mont mont)
 
 # ocsp_test
 if(ENABLE_EXTRATESTS)
+	if(NOT "${OPENSSLDIR}" STREQUAL "")
+		add_definitions(-D_PATH_SSL_CA_FILE=\"${OPENSSLDIR}/cert.pem\")
+	else()
+		add_definitions(-D_PATH_SSL_CA_FILE=\"${CMAKE_INSTALL_PREFIX}/etc/ssl/cert.pem\")
+	endif()
 	add_executable(ocsp_test ocsp_test.c)
 	target_link_libraries(ocsp_test ${OPENSSL_LIBS})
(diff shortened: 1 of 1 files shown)


busterb:
Nice catch, thanks for the patch.

tests/CMakeLists.txt
```diff
@@ -196,6 +196,11 @@ add_test(mont mont)
 
 # ocsp_test
 if(ENABLE_EXTRATESTS)
```
busterb: What about defining this as relative to CMAKE_CURRENT_SOURCE_DIR instead, so we don't assume that there are any certs in the install directory at all at test-time.

busterb:
This probably needs to be updated for automake too.

tests/CMakeLists.txt
geoffbeier: I think you're right and

# ocsp_test
if(ENABLE_EXTRATESTS)
	add_definitions(-D_PATH_SSL_CA_FILE=\"${CMAKE_CURRENT_SOURCE_DIR}/../apps/openssl/cert.pem\")
	add_executable(ocsp_test ocsp_test.c)
	target_link_libraries(ocsp_test ${OPENSSL_LIBS})
	add_test(ocsptest ${CMAKE_CURRENT_SOURCE_DIR}/ocsptest.sh)
	set_tests_properties(ocsptest PROPERTIES ENVIRONMENT "srcdir=${CMAKE_CURRENT_SOURCE_DIR}")
endif()

is actually better in principal, assuming add_definitions there doesn't have any effect outside the test directory. (I don't *think* it does, but I have not confirmed that.)

PR #227, opened 2016-09-18
export DLLs functions for MSVC with CMake

kinichiro:
- Add 3 DEF files to export functions from Windows DLLs
  - Add gettimeofday to crypto/crypto.def (*1)
  - Remove gai_strerrorA from tls/tls.def (*1)
- Fix CMakeLists.txt to use DEF files
- Change DLL import library file name since it duplicates with static library
- Ignore compiler warning C4267, and Edit CMAKE_C_FLAGS not to overwrite it (*1)
- Add USE_SHARED option to build openssl.exe with shared libraries (*1)
  
  (*1) recommended by @mcnameej


kinichiro:
I tested with CMake 3.6.1 and Visual Studio 2015.
To obtain Visual Studio Solution (LibreSSL.sln), run this command.

```
cmake -G "Visual Studio 14 2015 Win64"  ..
```

To build openssl.exe linking with DLLs, run this command.

```
cmake -G "Visual Studio 14 2015 Win64" -DUSE_SHARED=on  ..
```

workaround for undefined PATH_MAX, add definition in include/compat/limits.h

```
#ifndef PATH_MAX
#define PATH_MAX 260
#endif
```

mcnameej:
I'm confused about the PATH_MAX situation.  openssl.exe won't build on MSVC without PATH_MAX being defined, but your **fix_cmake_portable_2** branch doesn't include the limits.h change.

crypto/CMakeLists.txt
```diff
@@ -753,11 +753,14 @@ if (BUILD_SHARED)
 	add_library(crypto STATIC $<TARGET_OBJECTS:crypto-objects>)
 	add_library(crypto-shared SHARED $<TARGET_OBJECTS:crypto-objects>)
 	if (WIN32)
+		target_sources(crypto-shared PUBLIC
+			${CMAKE_CURRENT_SOURCE_DIR}/crypto.def)
```
mcnameej: Use PRIVATE, not PUBLIC.

busterb:
Commited as 62f2a73061eda53f5d60cfa7fab7dfb6bd7803ad, thanks!

PR #233, opened 2016-10-06
modify for Intel C++ Compiler

kinichiro:
- define _CRT_SUPPRESS_RESTRICT to avoid compilation error
- suppress compilation warnings (suggested by @Johnex)


busterb:
Committed as b434123987dc08fd47bd988bd3fe09277445001d

PR #235, opened 2016-10-20
Enable tests on Visual Studio

kinichiro:
- add patch for aeadtest.c to undef IN
- add patch for ocsp_test.c to call BIO_sock_init() before getaddrinfo()
- define STDERR_FILENO in unistd.h to build pkcs7test.c
- add option ENABLE_VSTEST(default OFF) to enable test on Visual Studio
- modify to pass test data file as an argument (aeadtest, evptest)
- add Windows scripts (ocsptest, pq_test, ssltest, testdsa, testenc, testrsa)
- do not build pidwraptest on MSVC
- fix some indentations


busterb:
Committed as 14905877a0eb85ebdc16162e820cda51f0895fc7

PR #236, opened 2016-10-20
remove unneeded slash after DESTDIR

kinichiro:


busterb:
Committed as 332c2eeaa3ad0133bd0ef725884652c509275c88

PR #242, opened 2016-11-05
implement support for hiding symbols in libtls / libssl

busterb:
This updates the Visual Studio DLL builds and autoconf builds to use explicitly-exported symbols.


kinichiro:
merged.

PR #248, opened 2016-12-09
configure: fix getentropy() for sierra and ios

bassosimone:
This diff changes the logic by which configure detects getentropy() to
ensure that we don't use the system wide getentropy

- with macOS sierra if the deployment target is lower than sierra as
  found by tor developers here

    https://gitweb.torproject.org/tor.git/commit/?id=https://gitweb.torproject.org/tor.git/commit/?id=16fcbd21c963a9a65bf55024680c8323c8b7175d

- with iOS unconditionally because an app linking libressl compiled with
  system wide getentropy has been rejected by the App store as I have
  documented here

    https://github.com/measurement-kit/measurement-kit/pull/994

I think something similar could also affect clock_gettime judging from
tor's patch, but this diff for now doesn't address that.

I do not have macOS < sierra, so I could only verify that configure was
not picking up system wide getentropy by compiling libressl using

    export CFLAGS="-mmacosx-version-min=10.11"

As regards iOS, removing the check for getentropy and recompiling (thus
using libressl builtin getentropy()) was enough to have another iteration
of the app accepted. Otherwise testing should be possible with:

    export LDFLAGS="-arch armv7 -miphoneos-version-min=7.1 -isysroot `xcrun --show-sdk-path --sdk iphoneos`"
    export CPPFLAGS="-arch armv7 -isysroot `xcrun --show-sdk-path --sdk iphoneos`"
    export CFLAGS="-arch armv7 -miphoneos-version-min=7.1 -isysroot `xcrun --show-sdk-path --sdk iphoneos`"

[edit: and by passing `--host=arm-apple-darwin` to configure]

Related ticket: https://github.com/libressl-portable/portable/issues/230

### DIFF
--- a/m4/check-libc.m4
+++ b/m4/check-libc.m4
@@ -52,10 +52,27 @@ AC_CHECK_FUNCS([explicit_bzero getauxval])
 AC_CACHE_CHECK([for getentropy], ac_cv_func_getentropy, [
 	AC_LINK_IFELSE([AC_LANG_PROGRAM([[
 #include <sys/types.h>
-#include <sys/random.h>
+#include <unistd.h>
 
 #ifdef __APPLE__
 #  include <AvailabilityMacros.h>
+#  include <TargetConditionals.h>
+
+# if (TARGET_OS_IPHONE || TARGET_OS_SIMULATOR)
+
+/*
+ * As of iOS 10.1, getentropy() as a system call is defined but is not
+ * declared in sys/random.h and submitting an App that links to getentropy()
+ * leads to the App store rejecting the App because:
+ *
+ * > The app references non-public symbols in $appname: _getentropy
+ *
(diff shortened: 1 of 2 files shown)


bassosimone:
I tested a206997 by compiling and checking the presence of symbols using `nm -g` in these cases:

- macos 10.12 with result: using system's arc4random_buf
- macos 10.12 with target 10.11 with result: both arc4random_buf and getentropy are builtin
- ios armv7 with result: both arc4random_buf and getentropy are builtin
- ios armv7s with result: both arc4random_buf and getentropy are builtin
- ios armv64 with result: both arc4random_buf and getentropy are builtin

I think my diff is now ready for review. Let me know if I should perform tests on other configurations or whether more changes are needed.

busterb:
Looks good to me, thanks!

PR #250, opened 2016-12-26
implement support for hiding symbols in libcrypto

kinichiro:
- delete crypto.def and generate it from Symbols.list
- remove BIO_s_log since bss_log.c is not included on Windows
- add Windows specific symbols by crypto/crypto_win.list


busterb:
Looks good, thanks @kinichiro.

PR #253, opened 2017-01-06
Add regression test ssl_versions

kinichiro:


4a6f656c:
Part of x509/x509_vfy.c -r1.54 has been reverted, which addresses the PKCS7_verify() regress failure.

kinichiro:
Thanks for fixing the issue.
I rerun the travis-ci and I could verify that build finished successfully.

PR #257, opened 2017-01-12
Fix to use -export-symbols for libcrypto

kinichiro:


busterb:
We need to explicitly export a few more symbols for portable too, note the travis failures:

```
  "_explicit_bzero", referenced from:
      _dtls1_free in d1_lib.o
      _ssl3_send_client_key_exchange in s3_clnt.o
      _ssl3_free in s3_lib.o
      _ssl3_clear in s3_lib.o
      _ssl3_get_client_key_exchange in s3_srvr.o
      _ssl3_send_newsession_ticket in s3_srvr.o
      _i2d_SSL_SESSION in ssl_asn1.o
      ...
  "_reallocarray", referenced from:
      _ssl_create_cipher_list in ssl_ciph.o
      _tls1_setup_key_block in t1_enc.o
      _ssl_parse_clienthello_tlsext in t1_lib.o
      _ssl_parse_serverhello_tlsext in t1_lib.o
  "_timingsafe_memcmp", referenced from:
```

kinichiro:
Yes, I noticed that we should modify the crypto.sym dynamically
while ./configure script running to export suitable symbols for each platforms.

OSX requires these symbols in crypto.sym.
explicit_bzero
reallocarray
timeingsafe_memcmp

Linux requires these symbols in crypto.sym.
arc4random_buf
explicit_bzero
reallocarray
strlcat
strlcpy

mingw 32 and 64 require BIO_s_log should be removed from crypto.sym.

kinichiro:
I will try to find another way to do this, and I closed this now.

PR #260, opened 2017-01-15
Fix to use -export-symbols for libcrypto

kinichiro:
- generating platform specific crypto/crypto_portable.sym from crypto.sym


busterb:
Looks better, thanks.

kinichiro:
Thanks for merging :-)

PR #263, opened 2017-01-24
initial ocspcheck integration

busterb:
This hasn't been tested yet except on OpenBSD, so failures expected during Travis CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,10 +117,15 @@ include/pqueue.h
 include/tls.h
 include/openssl/*.h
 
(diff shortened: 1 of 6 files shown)


kinichiro:
On Windows with Visual Studio 2015, strsep is missing in libssl.
libtls is using strsep() and it is included in libtls.
libssl can't use strsep() in libtls.
How do you think moving strsep() from libtls to libcrypto (compat) ?
Also adding prototype in include/compat/string.h.

busterb:
closing in favor of #264

PR #264, opened 2017-01-25
Add ocspcheck, move strsep to crypto/compat, add rsa_test

kinichiro:
This PR is addition for #263 of @busterb work.
I moved strsep.c from under tls/ to crypto/compat/.
I also added new regress rsa_test both autoconf and cmake.
This will work on Linux and OSX.
But on MinGW to build ocspcheck, we still need to resolve undefined references; inet_ntop(), S_IRGRP, __VA_ARGS__.
Additionally VisualStudio 2015, these are required; getopt(), ftruncate(), STDOUT_FILENO.


kinichiro:
Absolutely. In this PR, inet_ntop.c and memmem.c are copied everytime
from openbsd repository to apps/ocspcheck/compat/ by $CP_LIBC in update.sh.

```
$CP_LIBC $libc_src/net/inet_ntop.c apps/ocspcheck/compat
$CP_LIBC $libc_src/string/memmem.c apps/ocspcheck/compat
```

I had checked in these 2 files only for keeping apps/ocspcheck/compat/
directory in git repository.

Should we remove these 2 files from repository and put some empty file
in apps/ocspcheck/compat to keep this directory, for example,
apps/ocspcheck/compat/.gitignore ?
https://git.wiki.kernel.org/index.php/GitFaq#Can_I_add_empty_directories.3F

busterb:
Yeah, let's do that.

kinichiro:
I have done. Thanks for your advice.
We might do the same clean up for these directories below, but let's do that the other day, the other PR.
apps/nc/compat/
apps/openssl/compat/
crypto/compat/
tests/compat/

kinichiro:
I  should add define in include/compat/sys/stat.h 
rather than fcntl.h.
Do we need --build-ocspcheck switch like nc ?

busterb:
I don't think we need the check. Unlike nc, this is an entirely new command.

kinichiro:
I see.
Then we need to add getopt() and ftruncate() in apps/ocspcheck/compat/  for
Visual Studio.

busterb:
Roger. Hmm, let me think about that a bit!

busterb:
Rather than trying to whitelist for every OS like we do with nc, I think for now we could disable ocspcheck just for MSVC and then reenable in a separate PR once it's ready. I'll do that.

kinichiro:
Thanks. I think it is good choice for now.

PR #281, opened 2017-02-08
Fix CMake to generate DEF file with DATA keyword for exported data sy…

kinichiro:
…mbols

- For Windows DLL exported data symbols, add 'DATA' keyword as libtool does
- To detect data symbols from static library, using dumpbin.exe


kinichiro:
Hi,

This PR is experimental and I am appreciated any comments.

This affects for Windows Visual Studio CMake build only.
To export external variable (ex. *_it, SSL_version_str, etc.) in DLL, symbol should be followed by the keyword 'DATA' in .DEF file.
On Cygwin64 with dist-win.sh, libtool does the same kind of thing by using nm and sed.
So, I believe formal releases of LibreSSL Windows zip DLLs are build with DEF file includes 'DATA' keyword in it.
On Windows with CMake, we can't use nm and sed, then I tried to use dumpbin.exe and cmake script mode.

I verified this with VS2015, and CMake command is like below.

```
cmake -G "Visual Studio 14 2015 Win64" ..
```

With '-DUSE_SHARED=on', openssl.exe build failed with undefined reference NETSCAPE_X509_it, and I can't resolve this yet.

I wonder if this functionality is needed since I believe LibreSSL is going to reduce direct external variable access from outside of the library.

Best regards,

busterb:
I wonder if src/lib/libcrypto/asn1/x_nx509.c and NETSCAPE_X509_it can go away these days. Does anyone care about /* Old netscape certificate wrapper format */ ?

kinichiro:
Yeah, I don't like to merge this if no one care about that.
I will close this PR now.

busterb:
I suspect we could maybe fix the reference by removing the need for the symbol in the first place.

PR #286, opened 2017-02-14
Add -D__MINGW_USE_VC2005_COMPAT hint to avoid 32-bit time_t on mingw32

omega-software:
Hi
The proposed comment adds a very useful hint on how to solve the 32-bit time_t problem when compiling LibreSSL with Mingw-w64:

** On Mingw-w64, you can solve this problem by switching to
** MSVC 2005 ABI which provides 64-bit time_t. If you want to
** do that, rebuild $PACKAGE_NAME (and everything that links with it)
** with CPPFLAGS=-D__MINGW_USE_VC2005_COMPAT

PR #287, opened 2017-02-15
Document 32-bit time_t problem on minw-w64 toolchain and how to avoid it

omega-software:
Followup from https://github.com/libressl-portable/portable/pull/286

### DIFF
--- a/README.windows
+++ b/README.windows
@@ -12,7 +12,8 @@ cross compilers on Windows.
 To configure and build LibreSSL for a 32-bit system, use the following
 build steps:
 
- CC=i686-w64-mingw32-gcc ./configure --host=i686-w64-mingw32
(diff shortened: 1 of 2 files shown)


busterb:
Seems reasonable, I'll pull this in.

busterb:
Hmm, reviewing #286, do you feel this is worth documenting @kinichiro ? I feel that at least giving an option in the README (since most users on Win32 are likely going to to be building their entire stack themselves and statically linking) would be worthwhile.

kinichiro:
Yes, giving an option in the README would be worthwhile enough, and IMHO it is not needed in toolchain message.
Basically, I believe the information of this flag should be provided by README of application project that linking LibreSSL as library.
Since, all components are required to define this when the build time, and if not it will go wrong.
For the application projects manager, this info might be useful if this flag works fine with LibreSSL build on MinGW64.

PR #297, opened 2017-03-14
Add recallocarray

kinichiro:
This PR is initial support for using recallocarray.
We need to add getpagesize fix for VisualStudio and HP-UX.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,6 +142,7 @@ include/openssl/*.h
 !/crypto/compat/arc4random.h
 !/crypto/compat/b_win.c
 !/crypto/compat/explicit_bzero_win.c
+!/crypto/compat/getpagesize.c
 !/crypto/compat/posix_win.c
 !/crypto/compat/bsd_asprintf.c
 !/crypto/compat/inet_pton.c
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -681,6 +681,10 @@ if(NOT HAVE_ASPRINTF)
 	set(EXTRA_EXPORT ${EXTRA_EXPORT} vasprintf)
 endif()
 
+if(NOT HAVE_GETPAGESIZE)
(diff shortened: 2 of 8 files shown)


busterb:
Looks good to me. For Windows, we might use GetSystemInfo to get the page size.

kinichiro:
Yes, I agree to use GetSystemInfo.
I added 1 commit to this PR for supporting getpagesize.

- Add crypto/compat/getpagesize.c and if it is Windows call GetSystemInfo, other call sysconf(_SC_PAGESIZE)
- Checking if getpagesize exists by m4/check-libc.m4
- Do not export getpagesize since it is called only from libcrypto internal

First, I thought we could add portable patch only for recallocarray.c.
But I added this as internal function in case some other functions will call getpagesize in the future.
If portable patch method is better, I would change this.

BTW, in some man pages say that getpagesize is legacy and HP-UX does not have this.
Though I saw HP-UX 11.31 had it. Also Solaris 11.3 has it. AIX appears to have it, too.
So, every libressl supported unix like systems appear to have getpagesize at this time :-)
Then, if we don't need the part of sysconf(_SC_PAGESIZE) now, I would remove it.

busterb:
According to the linux manpage for getpagesize, it can sometimes lie and instead be static, depending on the architecture. The page doesn't seem to have been updated since about glibc 2.1 era, so it's hard to say how accurate this is today.

busterb:
Huh, Linux kernel's own documentation suggests using getpagesize(2) from userspace for determining ringbuffer sizes. I'm going to call that well supported.

busterb:
Yeah, let's leave out the compat for non-windows, then see what crops up.

kinichiro:
Yes, it is good to remove the compat code for non-windows, now. Thanks !

busterb:
Removing the code also highlighted a missing check in Cmake rules that was otherwise silent:  ff462f05a2e1337a5322eacb289e455be1d19e97

PR #301, opened 2017-04-10
move crypto_portable.sym to builddir

### LINKED ISSUE #300: crypto_portable.sym and out of tree build
Hi,
is out of tree build officially supported by libressl? If yes, one file was overlooked:

	$ CC=i686-w64-mingw32-gcc CPPFLAGS=-D__MINGW_USE_VC2005_COMPAT ../libressl-2.5.2/configure --prefix=~/mingw/i686-w64-mingw32 --host=i686-w64-mingw32
	$ git status
	On branch master
	Untracked files:
	  (use "git add <file>..." to include in what will be committed)

		libressl-2.5.2/crypto/crypto_portable.sym

	nothing added to commit but untracked files present (use "git add" to track)

busterb:
Generated files should be in the builddir rather than the srcdir. Because we're currently generating the file in configure's context, $builddir is not yet available. We should probably actually move the generation process to crypto/Makefile.am as well.

Fixes #300

### DIFF
--- a/crypto/Makefile.am
+++ b/crypto/Makefile.am
@@ -15,7 +15,10 @@ EXTRA_DIST += crypto.sym
 # needed for a CMake target
 EXTRA_DIST += compat/strcasecmp.c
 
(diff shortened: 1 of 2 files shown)


kinichiro:
It looks good to me.

busterb:
Thanks, I'll land it.

PR #311, opened 2017-05-06
Add definition of in_addr_t for Windows build

kinichiro:


kinichiro:
I'll come back here, again.
Now, I would close this.

PR #312, opened 2017-05-13
Add hkdf and Add in_addr_t for Windows

kinichiro:
- Add 1st step of hkdf function
- Add in_addr_t in include/compat/sys/types.h for Windows build


kinichiro:
Travis-ci tests failed but it was caused by tlstest.sh regression test.
Newly added hkdf regression test succeeded and Mingw build is ok.

busterb:
Looks good to me, thanks @kinichiro

PR #324, opened 2017-07-06
Add option LIBRESSL_SKIP_INSTALL

d3x0r:
Internally LIBRESSL_SKIP_INSTALL, if not set becomes ENABLE_LIBRESSL_INSTALL so this by default is enabled.  defining LIBRESSL_SKIP_INSTALL before hand will disable all install() rules.
This is useful if another project includes and links to this statically.
I chose to add a prefix to avoid potential name collision because the options are cached globally.

If the installation is skipped, maybe it should also disable building apps?  I didn't do that.


busterb:
Can you please update README.md with some documentation for this build option?

d3x0r:
Okay; rebased to fix conflicts.
Added readme.md documentation for all cmake options available.
Ended up making a lot of commits trying to get it to format correctly; the lines -DXXX=XX kept wanting to wrap with '-' on one line followed by the rest of the option, even in a code block using \```

d3x0r:
https://github.com/d3x0r/portable/blob/0e82f22d16dc4c45bd45f9a994a30fc7435d430b/README.md
link to how it will look after update; cmake options added at the very end.

busterb:
Great, thanks!

PR #325, opened 2017-07-06
update netcat patch

kinichiro:


busterb:
reading diffs of diffs is confusing, but this looks OK

PR #329, opened 2017-07-10
cmake_export_symbol: Darwin compatibility

frutiger:
The Darwin platform prefixes all C symbols with an underscore.  At
link-time of a shared library, libressl generates a list of symbols to
export, but does not prefix each symbol with an underscore.

This commit addresses that issue.


busterb:
We need to do this with automake too.

kinichiro:
I think libtool on OSX does the same kind of tweaks internally.
libtool seems try to hide the difference of exporting tricks between OSs.
I don't have OSX and can't see it, though,
Libtool on mingw64 was like that.

busterb:
I took a closer look, and agree @kinichiro. Thanks!

frutiger:
Thanks @kinichiro and @busterb.  I will double check why my build wasn't hitting `libtool`, and propose a reversion if this is not needed (or indeed if it breaks the *macOS* build).

frutiger:
I was able to verify that the change is needed when attempting to build shared libraries on *macOS*.  In particular this invocation: `cmake -DCMAKE_MACOSX_RPATH=on -DBUILD_SHARED=on -GNinja . && ninja` would otherwise lead to a failure.

PR #341, opened 2017-08-29
Add patch for gcm128.c to build with Visual Studio

kinichiro:
Hi,

I found regress aeadtest and gcm128test are failed on Windows Visual Studio.
And the commit https://github.com/libressl-portable/openbsd/commit/2005e957cabbfd8d5afc9bbc74ccf6144c448b0d relates to this issue.
By this commit, it detects the size of pointer with "#ifdef _LP64" instead of "sizeof(size_t)" on runtime.
Windows Visual Studio seems not to define "_LP64", and that brings wrong instruction execution.
To detect this on Windows Visual Studio, I think we can use "_WIN64".
I made a patch for this issue rather than requesting upstream modification since this issue is only for Windows.

Any comments are appreciated.

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -559,7 +559,6 @@ set(
 	rsa/rsa_pss.c
 	rsa/rsa_saos.c
 	rsa/rsa_sign.c
-	rsa/rsa_ssl.c
 	rsa/rsa_x931.c
 	sha/sha1_one.c
 	sha/sha1dgst.c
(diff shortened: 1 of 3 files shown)


kinichiro:
I simply added `add_definitions(-D_LP64)` in CMakeLists.txt and tried to build by VS2017.
But I had got a lot of compilation errors.
There are some other source files that uses `#ifdef _LP64`, and those might not to expect _LP64 == _WIN64.

kinichiro:
I have another idea to solve this.
Using SIZE_MAX in stdint.h like this,
`#if SIZE_MAX == 0xffffffffffffffff`
instead of
`#ifdef _LP64`

This way requires changing upstream, but I think this expression directly checks sizeof size_t is 8 or not.

busterb:
I agree. _LP64 is the wrong check, when we really wanted to test for pointer size = 64 bit.

kinichiro:
I will close this PR now.

PR #346, opened 2017-09-08
Some CMake Fixes.

lluixhi:
pqueue.h was getting installed when it shouldn't.
pkgconfig files were not getting installed or generated.

links to manual pages are also not being installed, but I'm not sure where to get a list of those links when I'm inside the tarball.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -324,6 +324,23 @@ if(NOT MSVC OR ENABLE_VSTEST)
 	add_subdirectory(tests)
 endif()
 
+if(NOT MSVC)
+	# Create pkgconfig files.
+	set(prefix      ${CMAKE_INSTALL_PREFIX})
+	set(exec_prefix \${prefix})
+	set(libdir      \${exec_prefix}/${CMAKE_INSTALL_LIBDIR})
+	set(includedir  \${prefix}/include)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -324,6 +324,23 @@ if(NOT MSVC OR ENABLE_VSTEST)
 	add_subdirectory(tests)
 endif()
 
+if(NOT MSVC)
+	# Create pkgconfig files.
+	set(prefix      ${CMAKE_INSTALL_PREFIX})
+	set(exec_prefix \${prefix})
+	set(libdir      ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR})
```
kinichiro: Hi,
This line should be `set(libdir \${exec_prefix}/lib`, no ?
Only libdir doesn't use variable to define its path.
Are there any reasons to use CMAKE_* variables ?

lluixhi: Whoops. No there isn't.

lluixhi: Keeping ${CMAKE_INSTALL_LIBDIR} though so it doesn't break lib64 systems.

kinichiro: Thanks for clarification. I understand. I think both resolutions for pkgconfig and pqueue.h are good.

PR #359, opened 2017-11-06
Create correct directory in CMake install.

lluixhi:
Just a minor gripe, but it doesn't match what's in Makefile.am and the source code.

cmake install creates ${CONFDIR}/cert instead of ${CONFDIR}/certs.


busterb:
Sure, thanks for the update.

PR #372, opened 2017-12-25
Add CLOCK_MONOTONIC and timersub for the OS that does not have them

kinichiro:
This Pull Request is for HP-UX build.

### DIFF
--- a/include/compat/sys/time.h
+++ b/include/compat/sys/time.h
@@ -13,4 +13,20 @@ int gettimeofday(struct timeval *tp, void *tzp);
 #include_next <sys/time.h>
 #endif
 
+#ifndef CLOCK_MONOTONIC
+#define CLOCK_MONOTONIC CLOCK_REALTIME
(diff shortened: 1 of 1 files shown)


include/compat/sys/time.h
```diff
@@ -13,4 +13,20 @@ int gettimeofday(struct timeval *tp, void *tzp);
 #include_next <sys/time.h>
 #endif
 
+#ifndef CLOCK_MONOTONIC
+#define CLOCK_MONOTONIC CLOCK_REALTIME
+#endif
+
+#ifndef timersub
+#define timersub(tvp, uvp, vvp)                                         \
+	do {                                                            \
```
heptathlon: Hey, what's the benefit of this do-while "loop"?

kinichiro:
In if or for block without bracket, this way of coding avoids unwanted
expansion.
This is a theory to write multi line macro.

PR #386, opened 2018-02-11
Add tls/tls_keypair.c

kinichiro:
- This PR solves linux and OSX build fail
- Added file permission define required by Windows and MinGW
- For Windows and MinGW build, need pread(), pwrite(), ftruncate() and gituid().

### DIFF
--- a/include/compat/sys/stat.h
+++ b/include/compat/sys/stat.h
@@ -10,8 +10,17 @@
 #include_next <sys/stat.h>
 
 /* for old MinGW */
+#ifndef S_IRWXU
+#define S_IRWXU         0
+#endif
+#ifndef S_IRWXG
+#define S_IRWXG         0
+#endif
 #ifndef S_IRGRP
 #define S_IRGRP         0
+#endif
+#ifndef S_IRWXO
+#define S_IRWXO         0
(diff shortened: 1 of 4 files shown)


RealDolos:
Is it deliberate this broke all mingw builds due to lack of `pread`, `pwrite`, `getuid`? Travis clearly complains, but then again you stated it would:

>  For Windows and MinGW build, need pread(), pwrite(), ftruncate() and gituid().

kinichiro:
Yes, I did. Before this PR, libtls newly required those system calls, and this PR mitigated the build break a little by adding some define like S_IRWXU and others.
After this we need to add pread(), pwrite(), ftruncate() and getuid() for MinGW and Windows.
For pread() and pwrite(), I think we could use read(), write() and lseek().
_chsize() would replace ftruncate().
Currently I have no idea about getuid() on Windows.
Could you help us for this issue, please ?

busterb:
Yeah, the latest is just what came out of the a2k18 hackathon.

We may need to look more high-level at what this is doing as well, since I believe that there may be some concepts like file descriptor passing that will need an entirely different approach on Windows (just substituting system calls might not be enough).

PR #398, opened 2018-03-14
Windows updates (drop XP/2003, add pthread_once)

busterb:
This change bumps our base OS requirements on Windows to Windows Vista (WIN32_WINTNT 0x600).

That means no more XP or 2003 support, which haven't been supported by Microsoft, technically, since 2014. It is not possible to even activate a new XP VM today, so I am unable to test builds.

Taking advantage of new APIs available under Vista, this drops the compat shims for inet_ntop/pton and adds a new pthread_once(3) shim for use in the updated initialization code.


busterb:
pthread_once(3) could be implemented with a lock instead if someone really cried out about XP/2003 compat. One of the slightly lame things about InitOnceExecuteOnce is that it can fail, which pthread_once(3) should not.

Maybe something like:
ftp://sourceware.org/pub/pthreads-win32/sources/pthreads-w32-2-9-1-release/pthread_once.c

bob-beck:
you are probably going to need pthread_self and pthread_equal soon too.
 apologies in advance.

kinichiro:
I tried to build on Windows7 with Visual Studio 2017.
And I found it still requires include/compat/arpa/inet.h.
Could you please remain compat/arpa/inet.h in include/Makefile.am ?
It also misses adding pthread.h.

4a6f656c:
I do not see dropping XP/2013 support as a bad thing (especially when it is untestable) - certainly not worth extra work IMO.

include/compat/pthread.h
```diff
@@ -0,0 +1,43 @@
...
+	INIT_ONCE once;
+};
+
+typedef struct pthread_once pthread_once_t;
+
+static BOOL CALLBACK _pthread_once_win32_cb(PINIT_ONCE once, PVOID param, PVOID *context)
+{
+	void (*cb) (void) = param;
+	cb();
+    return TRUE;
```
4a6f656c: Some strange indentation here...

include/compat/pthread.h
```diff
@@ -0,0 +1,43 @@
...
+}
+
+static int pthread_once(pthread_once_t *once, void (*cb) (void))
+{
+	BOOL rc = InitOnceExecuteOnce(&once->once, _pthread_once_win32_cb, cb, NULL);
+	if (rc == 0) {
+		return -1;
+	} else {
+		return 0;
+	}
```
kinichiro: These brackets are not needed since only 1 line statement under if and else.

bob-beck:
On Wed, Mar 14, 2018 at 12:40:05PM +0000, Brent Cook wrote:
> pthread_once(3) could be implemented with a lock instead if someone really cried out about XP/2003 compat. One of the slightly lame things about InitOnceExecuteOnce is that it can fail, which pthread_once(3) should not

Actually pthread_once is allowed to fail:

RETURN VALUES
     If successful, the pthread_once() function will return zero.  Otherwise
     an error number will be returned to indicate the error.

so failing if it doesn't work is theoretically fine.  Some code even
checks the return value :P

Bear in mind of course that much OpenSSL and OpenSSL using code
ignores errors on startup due to the crazy API, but people would
probably get mad at me if I sprinkled raise(SIGKILL) or raise(SIGABRT)
everywhere in the library to ensure it dies in these cases.

kinichiro:
I could verify that build on Visual Studio 2017 was OK.

I had found one issue.
Could you wrap regression keypairtest with "if(NOT MSVC)" in tests/CMakeLists.txt, please ?
This regression doesn't work well while buiding with VisualStudio.

Thanks.

PR #399, opened 2018-03-16
Use binary directory for symbol exports generation

donny-dont:
Currently the CMake files generate symbols into the source tree when they should go into the build tree.

I actually ran into this issue when attempting to make a [vcpkg](https://github.com/Microsoft/vcpkg) port of libressl. Internally vcpkg configures a debug and release build into different directories at the same time. Because the symbols files being generated were going into the same location within the source tree they were stepping on each other resulting in a garbled file.

Since these are generated file it seems like they should go into `${CMAKE_CURRENT_BINARY_DIR}` rather than within the source tree.


busterb:
Make sense, thanks for the catch.

PR #405, opened 2018-03-20
Add appveyor support for Window CI/CD

donny-dont:
LibreSSL does not have a Visual Studio build so its not clear if the CMake build works for it. This adds [AppVeyor](https://www.appveyor.com/) which is free for open source and has a Windows stack available for it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 environment:
   PATH: C:\msys64\usr\bin;C:\msys64\mingw64\bin;C:\Windows\System32;C:\Windows;%PATH%
-  VSTEST: OFF
+  VSTEST: ON
   
   matrix:
     # x86 builds
@@ -48,4 +48,4 @@ build_script:
   - cmake --build . --config %CONFIG%
 
 test_script:
-  - ctest
+  - ctest -C %CONFIG%
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,7 +15,7 @@ environment:
       SHARED_LIBS: ON
     - GENERATOR: Visual Studio 14 2015
       CONFIG: Debug
-      SHARED_LIBS: ON
+      SHARED_LIBS: OFF
 
     # x64 builds
     - GENERATOR: Visual Studio 14 2015 Win64
@@ -29,7 +29,7 @@ environment:
       SHARED_LIBS: ON
     - GENERATOR: Visual Studio 14 2015 Win64
       CONFIG: Debug
-      SHARED_LIBS: ON
+      SHARED_LIBS: OFF
 
 init:
   # update mysy2
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,51 @@
+environment:
+  PATH: C:\msys64\usr\bin;C:\msys64\mingw64\bin;C:\Windows\System32;C:\Windows;%PATH%
+  VSTEST: OFF
+  
+  matrix:
+    # x86 builds
+    - GENERATOR: Visual Studio 14 2015
+      CONFIG: Release
+      SHARED_LIBS: ON
+    - GENERATOR: Visual Studio 14 2015
+      CONFIG: Release
+      SHARED_LIBS: OFF
+    - GENERATOR: Visual Studio 14 2015
+      CONFIG: Debug
+      SHARED_LIBS: ON
+    - GENERATOR: Visual Studio 14 2015
+      CONFIG: Debug
+      SHARED_LIBS: ON
+
+    # x64 builds
+    - GENERATOR: Visual Studio 14 2015 Win64
+      CONFIG: Release
+      SHARED_LIBS: ON
+    - GENERATOR: Visual Studio 14 2015 Win64
+      CONFIG: Release
+      SHARED_LIBS: OFF
+    - GENERATOR: Visual Studio 14 2015 Win64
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -0,0 +1,51 @@
+environment:
+  PATH: C:\msys64\usr\bin;C:\msys64\mingw64\bin;C:\Windows\System32;C:\Windows;%PATH%
+  VSTEST: OFF
```
donny-dont: I turned `VSTEST` to `OFF` because I'm seeing [test failures](https://ci.appveyor.com/project/donny-dont/portable/build/1.0.54/job/uhcfotgb51cmf4v1) with it on and fixing tests is outside of scope for this.

Results of `ctest` are
```bash
89% tests passed, 7 tests failed out of 64
Total Test time (real) =   8.77 sec
The following tests FAILED:
	 41 - pq_test (Failed)
	 48 - servertest (Failed)
	 53 - ssltest (Failed)
	 54 - testdsa (Failed)
	 55 - testenc (Failed)
	 56 - testrsa (Failed)
	 59 - tlstest (Failed)
Errors while running CTest
Command exited with code 8
```

appveyor.yml
```diff
@@ -0,0 +1,51 @@
...
+    - GENERATOR: Visual Studio 14 2015 Win64
+      CONFIG: Debug
+      SHARED_LIBS: ON
+
+init:
+  # update mysy2
+  - C:\msys64\usr\bin\bash -lc "pacman --needed --noconfirm -Sy pacman-mirrors"
+  - C:\msys64\usr\bin\bash -lc "pacman --noconfirm -Sy"
+  - C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S autoconf perl"
+  - C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S git"
```
donny-dont: I couldn't get LibreSSL to build with the stock mingw that comes with its git. I saw that some projects actually updated msys64 so I went ahead and did that so the `autogen.sh` script could be run.

appveyor.yml
```diff
@@ -0,0 +1,51 @@
+environment:
+  PATH: C:\msys64\usr\bin;C:\msys64\mingw64\bin;C:\Windows\System32;C:\Windows;%PATH%
+  VSTEST: OFF
+  
+  matrix:
+    # x86 builds
+    - GENERATOR: Visual Studio 14 2015
+      CONFIG: Release
+      SHARED_LIBS: ON
+    - GENERATOR: Visual Studio 14 2015
```
donny-dont: I just did matrix builds of x86 and x64 builds. I'm not totally sure what version(s) of Visual Studio to support but the configs are all here.

busterb:
I just enabled AppVeyor for testing purposes. It might require a new commit to this branch for the output to appear.

busterb:
I noticed that AppVeyor can also generate artifacts from builds. It might be nice if the lib/dll/exe files got archived, so if someone wanted to grab them they're available automatically.

appveyor.yml
busterb: these appear to be all tests that have a batch file wrapper that drives an executable, so it might be a PATH or other issue with the test driver code.

appveyor.yml
```diff
@@ -0,0 +1,51 @@
...
+
+init:
+  # update mysy2
+  - C:\msys64\usr\bin\bash -lc "pacman --needed --noconfirm -Sy pacman-mirrors"
+  - C:\msys64\usr\bin\bash -lc "pacman --noconfirm -Sy"
+  - C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S autoconf perl"
+  - C:\msys64\usr\bin\bash -lc "pacman --noconfirm -S git"
+  
+before_build:
+  - bash autogen.sh
```
kinichiro: How about adding 
"make dist" and "tar zxf librssl-*.tar.gz" and cd there ?
Since this repository is tarball generator, and I think directly build here is not good.
I would like to test building the tarball rather than repository.

appveyor.yml
```diff
@@ -0,0 +1,51 @@
...
+    - GENERATOR: Visual Studio 14 2015
+      CONFIG: Release
+      SHARED_LIBS: OFF
+    - GENERATOR: Visual Studio 14 2015
+      CONFIG: Debug
+      SHARED_LIBS: ON
+    - GENERATOR: Visual Studio 14 2015
+      CONFIG: Debug
+      SHARED_LIBS: ON
+
```
kinichiro: This might be OFF.

appveyor.yml
kinichiro: Adding '.' (current directory) appears to solve that.

```
PATH: .;C:\msys64\usr\bin;C:\msys64\mingw64\bin;C:\Windows\System32;C:\Windows;%PATH%
```

kinichiro: That's why PATH 'Debug' is used in batch files.
In those days, I had no idea to pass build environment 'Release' or 'Debug' to test driver batch files.
So, with Debug build, regressions will be succeeded now, but with Release would be failed.

appveyor.yml
busterb: This also needs to be added to the mingw test builds in travis as well. We'll get it in a later PR.

busterb:
Thanks @donny-dont , this is pretty handy to have!

PR #406, opened 2018-03-20
Use BUILD_SHARED_LIBS to specify library type

donny-dont:
Removes custom logic for shared libraries and instead uses [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/v3.11/variable/BUILD_SHARED_LIBS.html) which is available in 2.8.8.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,23 +262,7 @@ if(ENABLE_ASM)
 	endif()
 endif()
 
-if(NOT (CMAKE_SYSTEM_NAME MATCHES "(Darwin|CYGWIN)"))
-	set(BUILD_SHARED true)
-endif()
-
-# USE_SHARED builds applications (e.g. openssl) using shared LibreSSL.
-# By default, applications use LibreSSL static library to avoid dependencies.
-# USE_SHARED isn't set by default; use -DUSE_SHARED=ON with CMake to enable.
-# Can be helpful for debugging; don't use for public releases.
-if(NOT BUILD_SHARED)
-	set(USE_SHARED off)
-endif()
-
-if(USE_SHARED)
-	set(OPENSSL_LIBS tls-shared ssl-shared crypto-shared)
-else()
-	set(OPENSSL_LIBS tls ssl crypto)
-endif()
+set(OPENSSL_LIBS tls ssl crypto)
 
 if(WIN32)
 	set(OPENSSL_LIBS ${OPENSSL_LIBS} ws2_32)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,10 +29,10 @@ string(REGEX REPLACE "\\..*" "" TLS_MAJOR_VERSION ${TLS_VERSION})
 
 option(LIBRESSL_SKIP_INSTALL "Skip installation" ${LIBRESSL_SKIP_INSTALL})
 option(LIBRESSL_APPS "Build apps" ON)
+option(LIBRESSL_TESTS "Build tests" ON)
 option(ENABLE_ASM "Enable assembly" ON)
 option(ENABLE_EXTRATESTS "Enable extra tests that may be unreliable on some platforms" OFF)
 option(ENABLE_NC "Enable installing TLS-enabled nc(1)" OFF)
-option(ENABLE_VSTEST "Enable test on Visual Studio" OFF)
 set(OPENSSLDIR ${OPENSSLDIR} CACHE PATH "Set the default openssl directory" FORCE)
 
 if(NOT LIBRESSL_SKIP_INSTALL)
@@ -300,7 +300,8 @@ add_subdirectory(include)
 if(NOT MSVC)
 	add_subdirectory(man)
 endif()
-if(NOT MSVC OR ENABLE_VSTEST)
+# Tests require the openssl executable and are unavailable when building shared libraries
+if(LIBRESSL_APPS AND LIBRESSL_TESTS AND NOT BUILD_SHARED_LIBS)
 	add_subdirectory(tests)
 endif()
 
--- a/README.md
+++ b/README.md
@@ -143,9 +143,11 @@ into other projects or build by itself.
 | Option Name | Default | Description
 | ------------ | -----: | ------
 |  LIBRESSL_SKIP_INSTALL | OFF | allows skipping install() rules.  Can be specified from command line using <br>```-DLIBRESSL_SKIP_INSTALL=ON``` |
+|  LIBRESSL_APPS | ON | allows skipping application builds. Apps are required to run tests |
+|  LIBRESSL_TESTS | ON | allows skipping of tests. Tests are only available in static builds |
+|  BUILD_SHARED_LIBS | OFF | CMake option for building shared libraries. |
 |  ENABLE_ASM | ON | builds assembly optimized rules. |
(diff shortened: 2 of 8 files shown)


donny-dont:
@busterb I have a Appveyor build with the contents of this patch in https://ci.appveyor.com/project/donny-dont/portable/build/1.0.59 to show it compiles and works as expected.

busterb:
This causes a regression with OpenBSD builds at least. I'll try to enable the tests in master for travis so we can see our baseline.

```
$ ninja                                                                       
[321/384] Linking C executable tests/tls_prf
FAILED: tests/tls_prf 
: && /usr/bin/cc    tests/CMakeFiles/tls_prf.dir/tls_prf.c.o  -o tests/tls_prf -L/home/bcook/projects/libressl/portable/build/tls  -L/home/bcook/projects/libressl/portable/build/ssl  -L/home/bcook/projects/libressl/portable/build/crypto -Wl,-rpath,/home/bcook/projects/libressl/portable/build/tls:/home/bcook/projects/libressl/portable/build/ssl:/home/bcook/projects/libressl/portable/build/crypto -ltls -lssl -lcrypto -Wl,-rpath-link,/usr/X11R6/lib:/usr/local/lib && :
tests/CMakeFiles/tls_prf.dir/tls_prf.c.o: In function `do_tls_prf_test':
../tests/tls_prf.c:(.text+0x17d): undefined reference to `ssl3_get_cipher_by_value'
/usr/bin/ld: tests/CMakeFiles/tls_prf.dir/tls_prf.c.o: relocation R_X86_64_PC32 against `ssl3_get_cipher_by_value' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
cc: error: linker command failed with exit code 1 (use -v to see invocation)
[323/384] Linking C executable tests/cipher_list
FAILED: tests/cipher_list 
: && /usr/bin/cc    tests/CMakeFiles/cipher_list.dir/cipher_list.c.o  -o tests/cipher_list -L/home/bcook/projects/libressl/portable/build/tls  -L/home/bcook/projects/libressl/portable/build/ssl  -L/home/bcook/projects/libressl/portable/build/crypto -Wl,-rpath,/home/bcook/projects/libressl/portable/build/tls:/home/bcook/projects/libressl/portable/build/ssl:/home/bcook/projects/libressl/portable/build/crypto -ltls -lssl -lcrypto -Wl,-rpath-link,/usr/X11R6/lib:/usr/local/lib && :
tests/CMakeFiles/cipher_list.dir/cipher_list.c.o: In function `ssl_bytes_to_list_alloc':
../tests/cipher_list.c:(.text+0x1f2): undefined reference to `CBS_init'
/usr/bin/ld: tests/CMakeFiles/cipher_list.dir/cipher_list.c.o: relocation R_X86_64_PC32 against `CBS_init' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
cc: error: linker command failed with exit code 1 (use -v to see invocation)
[326/384] Linking C executable tests/pkcs7test
ninja: build stopped: subcommand failed.
```

donny-dont:
@busterb how is it being built? What is the command line for cmake?

donny-dont:
Adding a commit to make travis attempt to do a shared build with CMake to see if this shows anything.

donny-dont:
Travis build fails when you build out shared libs. Previously with tests in CMake they would always link to the static versions. I'm guessing that there are internal functions within the libraries that are not exposed which is causing the linker errors.

That sound correct @busterb ?

CMakeLists.txt
```diff
@@ -316,7 +300,8 @@ add_subdirectory(include)
 if(NOT MSVC)
 	add_subdirectory(man)
 endif()
-if(NOT MSVC OR ENABLE_VSTEST)
+# Tests require the openssl executable and are unavailable when building shared libraries
+if(LIBRESSL_APPS AND LIBRESSL_TESTS AND NOT BUILD_SHARED_LIBS)
```
donny-dont: It appears that in the `.sym` files that certain functions are internal and not exposed in the .so. Because of this tests are disabled on `BUILD_SHARED_LIBS=ON`. I think this is the root cause of why the tests were failing on OpenBSD.

tests/CMakeLists.txt
```diff
@@ -12,155 +12,143 @@ include_directories(
 
 add_definitions(-D_PATH_SSL_CA_FILE=\"${CMAKE_CURRENT_SOURCE_DIR}/../apps/openssl/cert.pem\")
 
-foreach(lib IN LISTS OPENSSL_LIBS)
-	if(${lib} STREQUAL "tls-shared")
```
donny-dont: This block implies that the tests will not build with shared libraries.

scripts/travis
```diff
@@ -12,20 +12,32 @@ if [ "x$ARCH" = "xnative" ]; then
...
 		sudo apt-get update
 		sudo apt-get install -y cmake ninja-build
 		cmake -GNinja ..
 		ninja
 		ninja test
+
+		cd ../build-shared
+		cmake -GNinja -DBUILD_SHARED_LIBS=ON ..
+		ninja
+		ninja test
```
donny-dont: This and the above `make test` can probably be removed though I don't expect any issues here. I'm not sure if at some point in the future it will be possible to build and run tests in shared mode.

README.md
```diff
@@ -143,9 +143,11 @@ into other projects or build by itself.
 | Option Name | Default | Description
 | ------------ | -----: | ------
 |  LIBRESSL_SKIP_INSTALL | OFF | allows skipping install() rules.  Can be specified from command line using <br>```-DLIBRESSL_SKIP_INSTALL=ON``` |
+|  LIBRESSL_APPS | ON | allows skipping application builds. Apps are required to run tests |
+|  LIBRESSL_TESTS | ON | allows skipping of tests. Tests are only available in static builds |
+|  BUILD_SHARED_LIBS | OFF | CMake option for building shared libraries. |
```
donny-dont: Let me know if more documentation is required for this.

kinichiro:
Hi,

LibreSSL hides non-public internal API from shared library.
Application program should use exported public API only.
But, some regression tests are using those hiding internal functions for
test purpose.
To build those tests, linking with static library is required since it
provides internal too.

busterb:
Seems like the behavior is still a bit different than before. Can this build shared and static in a single go?

kinichiro:
I would like to keep running regressions with BUILD_SHARED_LIBS = ON,
even if some of them require the static library to build.
This PR prohibits building tests while BUILD_SHARED_LIBS = ON.

donny-dont:
@busterb @kinichiro the behavior is different. Its a CMakeism to use `BUILD_SHARED_LIBS` to specify that all calls to add_library within a project should be built as `STATIC` or `SHARED`. If someone wanted to build both they would have to invoke CMake twice one with `-DBUILD_SHARED_LIBS=ON`.

When using `vcpkg` it wants that sort of layout and pushes ports towards using `BUILD_SHARED_LIBS` since it uses a different triplet `x64-windows-static` for static builds. I can just patch locally if this isn't what you guys want.

busterb:
What do you think @kinichiro - sounds like just invoking twice will work.

kinichiro:
@busterb ,
I would feel like to build regression tests against to shared libraries,
and run them.
Not only against to static libraries.
I thought this PR couldn't this.
Or do I miss anything here ?
Can I test with shared ssl, crypto and libtls libraries if I run the build
twice ?

PR #432, opened 2018-07-20
Generate regress freenull.c

kinichiro:


botovq:
This looks good to me.

ok tb

PR #439, opened 2018-08-14
Add portable scaffold for SM3

randombit:
This is the companion to https://github.com/libressl-portable/openbsd/pull/98 adding support for SM3 hash function.

CC @ronaldtse


busterb:
Thanks, taking a look now that it is upstream.

Issue #446, opened 2018-09-28
Missing links


diorcety:
Using cmake the library generated doesn't link to pthread and rt:

```
ldd lib/libssl.so
	linux-vdso.so.1 (0x00007ffdd957c000)
	libcrypto.so.43 => not found
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2942ab1000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2942e50000)
```

```
ldd lib/libcrypto.so
	linux-vdso.so.1 (0x00007ffc82586000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8a02ae7000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f8a02eb2000)
```

You should add target_link_libraries(target ${PLATFORM_LDADD})

kinichiro:
I have some questions.

1. What OS and version do you use ?

2. After running cmake, can you `make` and `make test` ?
   If you can `make test` and all regression tests running, there is no problem.

diorcety:
I will do that tomorrow. I'm on a linux système. It's easily to reproduice. Try to compile libarchive with liressl. You have to add -lrt and -pthread, to correctly built the binaires.

diorcety:
With cmake
```
./crypto/libcrypto.so:
	linux-vdso.so.1 (0x00007fff317ed000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f6895a49000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f689620a000)
undefined symbol: pthread_once	(./crypto/libcrypto.so)
./tls/libtls.so:
	linux-vdso.so.1 (0x00007fffee91f000)
	libssl.so.45 => /home/tt/tmp/libressl/build/ssl/libssl.so.45 (0x00007f7447861000)
	libcrypto.so.43 => /home/tt/tmp/libressl/build/crypto/libcrypto.so.43 (0x00007f744743f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f74470a0000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f7447cd3000)
undefined symbol: pthread_once	(/home/tt/tmp/libressl/build/crypto/libcrypto.so.43)
undefined symbol: pthread_once	(/home/tt/tmp/libressl/build/ssl/libssl.so.45)
[...]

PR #447, opened 2018-10-03
Fix cmake linking libraries

kinichiro:
Tweak platform specific linking libraries in cmake as suggested by issue #446 .


shinchiro:
The .pc generated doesn't append '-l' before the lib. This is example of pc file generated by cmake

```
#libcrypto pkg-config source file

prefix=/build/install/mingw
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: LibreSSL-libcrypto
Description: LibreSSL cryptography library
Version: 2.8.1
Requires:
Conflicts:
Libs: -L${libdir} -lcrypto
Libs.private:  ws2_32
Cflags: -I${includedir}
```

kinichiro:
I had added another commit to fix Libs.private in .pc files by adding "-l".
Could you check this, please ?

shinchiro:
Looks right 👍 The generated .pc files are similar with what makefile would produce

kinichiro:
Thanks for merging :)

PR #453, opened 2018-11-08
Misc fixes to bring portable in line with upstream.

bobsayshilol:
Without these changes `update.sh` fails to complete due to missing files and failed patches. This was a quick fix as I wanted to check something so I may have missed something or not done something the proper way, but `make; make test;` succeeds for me as does linking to my project so I may as well push this to save anyone else the trouble.


busterb:
Thanks mate, I was doing the same thing :) I'll merge and reconcile our diffs.

bobsayshilol:
Thanks! Looking at the subsequent commits I guess I did miss something, oops.

PR #475, opened 2018-12-14
Fix CMake not finding getpagesize for Android 19

### LINKED ISSUE #471: Android build failure with CMake
Hello,

For Android versions prior to API Level 21, `getpagesize` is defined inline in `unistd.h` and is not detected properly by the CMakeLists.txt, resulting in a re-definition that breaks the build.

Currently LibreSSL looks for functions using [`check_function_exists`](https://cmake.org/cmake/help/v3.13/module/CheckFunctionExists.html), but as the doc mentions it can't find inline functions and should be replaced by [`check_symbol_exists`](https://cmake.org/cmake/help/v3.13/module/CheckSymbolExists.html#module:CheckSymbolExists)

This issue seems to be similar to #354, but this time the problem is with CMake.
I can confirm that replacing just the 
`check_function_exists(getpagesize HAVE_GETPAGESIZE)` 
line by
`check_symbol_exists(getpagesize unistd.h HAVE_GETPAGESIZE)`
fixes the issue, but ideally all the `check_function_exists` calls should be replaced.

tux3:
Using `check_symbol_exists` CMake is now able to find the inline getpagesize function on Android API 19.

Note that CMake 3.0 is required for this and the CMakeLists currently asks for 2.8.8, but since some `target_*` APIs are in use 2.8.8 support was already broken in practice.

This should fix #471 (ping @busterb).


busterb:
Looks good to me, fine with bumping cmake version.

AliceWonderMiscreations:
Will this be android specific or in general?

Note that some enterprise Linux distributions (e.g. CentOS 7) still package cmake 2.8.x.

That being said I personally already use a newer cmake in my own package build system for CentOS 7 because a few other projects I build also need >= 3 so it isn't an issue for me, but it might generate some build failure from others who do not already have an updated cmake on their build system.

busterb:
@AliceWonderMiscreations as @tux3 pointed out, we were already accidentally requiring 3.0 or later anyway.

busterb:
According to #482 this seems to still be an issue. I wonder if this needs different detection on different Android API versions. What a pain if so.

PR #485, opened 2019-01-01
adjust Windows locks for auto-initialization

### LINKED ISSUE #478: Function crypto_init_locks not being called
While working on this pull request: https://github.com/libressl-portable/portable/pull/477
I saw that crypto_init_locks is not being called automatically anywhere in the library. That means that the WaitForSingleObject functions are using un-initialized variables, and therefore firing instantly, without waiting for a lock.

The locks are at the same time not freed anywhere in the library either, leaking memory (CloseHandle not being called) on program close.

busterb:
This is a modification of #477 from @johnex that uses an atomic swap to perform initialization as-needed. This is the same as how arc4random lock initialization works in arc4random_win.h

Fixes #478

### DIFF
--- a/crypto/compat/crypto_lock_win.c
+++ b/crypto/compat/crypto_lock_win.c
@@ -19,16 +19,7 @@
 
 #include <openssl/crypto.h>
 
-static HANDLE locks[CRYPTO_NUM_LOCKS];
-
-void
-crypto_init_locks(void)
-{
-	int i;
-
-	for (i = 0; i < CRYPTO_NUM_LOCKS; i++)
-		locks[i] = CreateMutex(NULL, FALSE, NULL);
-}
+static volatile LPCRITICAL_SECTION locks[CRYPTO_NUM_LOCKS] = { NULL };
 
 void
 CRYPTO_lock(int mode, int type, const char *file, int line)
@@ -36,10 +27,20 @@ CRYPTO_lock(int mode, int type, const char *file, int line)
 	if (type < 0 || type >= CRYPTO_NUM_LOCKS)
 		return;
 
+	if (locks[type] == NULL) {
+		LPCRITICAL_SECTION lcs = malloc(sizeof(CRITICAL_SECTION));
+		InitializeCriticalSection(lcs);
+		if (InterlockedCompareExchangePointer((void **)&locks[type], (void *)lcs, NULL) != NULL)
+		{
+			DeleteCriticalSection(lcs);
+			free(lcs);
+		}
+	}
+
 	if (mode & CRYPTO_LOCK)
-		WaitForSingleObject(locks[type], INFINITE);
+		EnterCriticalSection(locks[type]);
 	else
-		ReleaseMutex(locks[type]);
+		LeaveCriticalSection(locks[type]);
 }
 
 int
--- a/crypto/compat/crypto_lock_win.c
+++ b/crypto/compat/crypto_lock_win.c
@@ -30,8 +30,7 @@ CRYPTO_lock(int mode, int type, const char *file, int line)
 	if (locks[type] == NULL) {
 		LPCRITICAL_SECTION lcs = malloc(sizeof(CRITICAL_SECTION));
 		InitializeCriticalSection(lcs);
-		if (InterlockedCompareExchangePointer((void **)&locks[type], (void *)lcs, NULL) != NULL)
-		{
+		if (InterlockedCompareExchangePointer((void **)&locks[type], (void *)lcs, NULL) != NULL) {
 			DeleteCriticalSection(lcs);
 			free(lcs);
 		}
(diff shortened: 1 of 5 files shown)


busterb:
Hmm, I initially tested, but forgot to actually compile it apparently... not sure why I thought 0ce4911 was necessary, but will migrate other files to avoid using that pattern, since it's an easy way to build the wrong code if you're not careful.

busterb:
Thanks @Johnex . The solution above appears to swap a stack pointer over the locks array, which probably isn't the right thing. To support this mode of atomic initialization, we probably need to have an array of `LPCRITICAL_SECTION`, and allocate the pointer if we continue using InterlockedCompareExchangePointer to perform per-lock initialization on the fly.

In the mean time, I updated this diff to use a Mutex for now, to get the initialization working correctly, which is now matching the pattern we use in arc4random_win.h. When we find a good initialization pattern with CRITICAL_SECTION, it probably makes sense to switch arc4random_win.h as well.

Johnex:
Yup i am testing at the moment some fixes. The code above deadlocks after just the compile fixes.

Johnex:
```c
/* $OpenBSD: crypto_lock.c,v 1.1 2018/11/11 06:41:28 bcook Exp $ */
/*
 * Copyright (c) 2018 Brent Cook <bcook@openbsd.org>
 *
 * Permission to use, copy, modify, and distribute this software for any
 * purpose with or without fee is hereby granted, provided that the above
 * copyright notice and this permission notice appear in all copies.
 *
 * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
 * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
 * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
 * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
 * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
 * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
 * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 */

#include <windows.h>

#include <openssl/crypto.h>

static volatile LPCRITICAL_SECTION locks[CRYPTO_NUM_LOCKS] = { NULL };

void
CRYPTO_lock(int mode, int type, const char *file, int line)
{
	if (type < 0 || type >= CRYPTO_NUM_LOCKS)
		return;

	if (locks[type] == NULL) {
		LPCRITICAL_SECTION lcs = malloc(sizeof(CRITICAL_SECTION));
		InitializeCriticalSection(lcs);
		if (InterlockedCompareExchangePointer(&locks[type], lcs, NULL) != NULL)
		{
			DeleteCriticalSection(lcs);
			free(lcs);
		}
	}

	if (mode & CRYPTO_LOCK)
		EnterCriticalSection(locks[type]);
	else
		LeaveCriticalSection(locks[type]);
}

int
CRYPTO_add_lock(int *pointer, int amount, int type, const char *file,
    int line)
{
	int ret = 0;
	CRYPTO_lock(CRYPTO_LOCK|CRYPTO_WRITE, type, file, line);
	ret = *pointer + amount;
	*pointer = ret;
	CRYPTO_lock(CRYPTO_UNLOCK|CRYPTO_WRITE, type, file, line);
	return (ret);
}
```

Now works as it should. Some notes.
The memory claimed by malloc will be freed automatically since InterlockedCompareExchangePointer provides a full memory barrier.
If 2 threads pass the if (locks[type] == NULL) at the same time, InterlockedCompareExchangePointer is it's own mutex and will only let 1 thread exchange the pointer, gracefully failing and freeing the memory and critical section.

I think this is a good implementation, memory will be freed and the only extra slowdown is the NULL check each time CRYPTO_lock is called, but we gain the automatic initialization part.

busterb:
Looks good, I'll update and rerun tests with this.

busterb:
I added some additional casts that were needed when building with mingw-w64, and switched CRYTPO_add_lock to use InterlockedExchangeAdd. I think that'll work a bit better than doing a lock acquire/release as well. Now to let the tests run, but I think this'll work well.

Johnex:
Looks good and cleaner too 👍 😀

crypto/compat/crypto_lock_win.c
```diff
@@ -19,37 +19,36 @@
...
+static volatile LPCRITICAL_SECTION locks[CRYPTO_NUM_LOCKS] = { NULL };
 
 void
 CRYPTO_lock(int mode, int type, const char *file, int line)
 {
 	if (type < 0 || type >= CRYPTO_NUM_LOCKS)
 		return;
 
+	if (locks[type] == NULL) {
+		LPCRITICAL_SECTION lcs = malloc(sizeof(CRITICAL_SECTION));
```
4a6f656c: The return value of `malloc()` appears to be unchecked.

Issue #482, opened 2018-12-23
Help building android


d3x0r:
the fix for getpagesize doesn't work (2.9.0)  Still have to do some hack around that.
Also ./include/compat/machine/endian.h  conflicts with /usr/include/machine.h because the include in the system is done with "machine/endian.h"... so I have to opt out

__ANDROID__ is my cmake build symbol for being android build...

```
if( NOT __ANDROID__ )
  set( more_include_directories ../include/compat )
endif( NOT __ANDROID__ )

add_library(tls ${TLS_SRC})
target_include_directories(tls
	PRIVATE
		.
		${more_include_directories}
	PUBLIC
		../include)

```

and make the above modification in crypto/CMakeLists.txt, tls/CMakeLists.txt and ssl/CMakeLists.txt

---

and that leaves me at 

```
[ 45%] Building C object src/contrib/libressl/2.9.0/crypto/CMakeFiles/crypto.dir/compat/getpagesize.c.o
M:\sack\src\contrib\libressl\2.9.0\crypto\compat\getpagesize.c:10:1: error: redefinition of 'getpagesize'
 getpagesize(void) {
 ^
In file included from M:\sack\src\contrib\libressl\2.9.0\crypto\compat\getpagesize.c:3:0:
C:/general/android/android-ndk-r9d/platforms/android-14/arch-arm/usr/include/unistd.h:171:23: note: previous definition of 'getpagesize' was here
 static __inline__ int getpagesize(void) {
                       ^
```

---

And in crytpo/modes/modes_lcl.h Added __ARM_ARCH > 5 (current build was 5, and it was failing, don't know the exact version to get 'rev' assembly opcode....

```
# elif (defined(__arm__) || defined(__arm)) && !defined(__STRICT_ALIGNMENT) && ( __ARM_ARCH > 5 )

```

busterb:
What Android API version are you building to? Maybe @tux3 can help you out.

busterb:
Oh, nevermind. API version 14 it looks like.

busterb:
I'll see about getting the SIZE_MAX changes in upstream, thanks.

busterb:
See #487 for a possible workaround for SIZE_MAX, though I haven't done anything more than a local compile with it.

busterb:
[...]

PR #487, opened 2019-01-01
add stdint.h shim, include limits for SIZE_MAX on Android

busterb:
This adds a workaround shim for some Android SDKs where SIZE_MAX is defined in limits.h instead of stdint.h.

See #482, this is a speculative diff, I haven't tested it.


busterb:
Tested with a few different SDKs, seems to do the trick.

PR #488, opened 2019-01-02
Android build fixes

busterb:
This fixes a few issues building out of the box using an Android NDK building against Bionic. See #482 for more details. A bigger change here is automatically determining if we need to link libpthread; this will need some work and testing, but I think will be a good change from hardcoding it for platforms.

This also probably needs some tweaks for CMake


busterb:
@kinichiro would you be able to verify that this works as expected on HP-UX?

PR #491, opened 2019-01-03
Add ARM build

kinichiro:
This PR adds ARM32 and ARM64 build by Travis-CI.
Regression tests are executed under QEMU.
Include ARM header file within distribution package.


busterb:
Thanks, I had a local diff for the header addition, but the travis build is nice!

PR #492, opened 2019-01-04
Handle malloc returning null

Johnex:
Locks are required for multi-threading. If memory can't be allocated, exit the program with memory error. If we let the program continue, it will deadlock in the next part of the code anyway so better end it before.


busterb:
Looks like the appveyor failure was due to networking, nothing in this PR.

Johnex:
I saw some networking issues in your pulls too. Is there a way to make appveyor retry a few times on network error?

busterb:
Probably not worth the effort. I can click a button to retry the build or any sub-task, but since it only failed on one, I wasn't worried much about it.

PR #503, opened 2019-02-11
Add LibreSSL CMake Find Module and update Readme

Johnex:


snikulov:
@Johnex It is better to add support to package config, IMO.
Find module has a little chance to be included into CMake, while package config is a preferred method to configure the external projects.

Johnex:
I don't want it to be included into CMake because they like to strip the licenses from the files and makes all code appear like their own, unless you check the git. This will just be included into LibreSSL as a convenient way to load the library.

snikulov:
@Johnex I don't take in count that your main intention was a license with your name in it for such simple script. I'm really sorry.

Johnex:
It might be simple, but there is no module currently for LibreSSL so its nice to have :)

snikulov:
> I know its possible to make a support package that is generated on install, but since there are multiple ways to compile this module, make + autoconf the most popular for linux, this is simpler.

For make + autoconf better to use pkg-config 
CMake also have support for it.

Johnex:
You are free to make a new pull request that you think is better 😄

Issue #507, opened 2019-03-04
Build Failed: missing t1_hash.c after running ./autogen.sh?


arun-ajay:
I ran the following on the terminal (Ubuntu 18.04):

git clone 'https://github.com/libressl-portable/portable'
cd portable/
./autogen.sh

Here is the output and error given:
```
pulling upstream openbsd source
Cloning into 'openbsd'...
remote: Enumerating objects: 1831, done.
remote: Counting objects: 100% (1831/1831), done.
remote: Compressing objects: 100% (636/636), done.
remote: Total 99161 (delta 1048), reused 1551 (delta 959), pack-reused 97330
Receiving objects: 100% (99161/99161), 32.02 MiB | 21.87 MiB/s, done.
Resolving deltas: 100% (72005/72005), done.
Already on 'master'
Your branch is up to date with 'origin/master'.
Already up to date.
Current branch master is up to date.
libcrypto version 45:3:0
libssl version 47:3:0
libtls version 19:4:0
Added OID sm3
Added OID sm3WithRSAEncryption
LibreSSL version 2.9.0
copying libcrypto source
generating arm ASM source for elf
generating x86_64 ASM source for elf
generating x86_64 ASM source for macosx
generating x86_64 ASM source for masm
generating x86_64 ASM source for mingw64
copying libtls source
copying nc(1) source
copying ocspcheck(1) source
copying openssl(1) source
copying libssl source
cp: cannot stat '/home/user/portable/openbsd/src/lib/libssl/t1_hash.c': No such file or directory
```

I can't locate t1_hash.c from the source. Where is it?

busterb:
Fixed in #508

PR #508, opened 2019-03-04
Add ssl/ssl_transcript.c and remove ssl/t1_hash.c

kinichiro:
This PR fixes latest build error and solve #507 .
For Windows build failure C2036, another bits are required.


botovq:
This looks good to me.  The appveyor breakage is addressed by millert's
const commit, isn't it?

In that case, I think you should merge this and close #509. 

Thanks

busterb:
Got it, merging and doing a source update on the openbsd upstream tree now.

Issue #474, opened 2018-12-11
AES-NI: Enable it in Windows platform


lars18th:
Hi,

Any idea why the AES-NI support is enabled for Linux x64 and MacOSX 64 bit builds, and not for Windows 64bit target?

I see the config files: `CMakeLists.txt`, `Makefile.am.elf-x86_64` and `Makefile.am.macosx-x86_64` with directives to compile `aes/aesni*.S` files. However, in the target repository of OpenSSL identical files `aes/aes*.asm` exist. Check it:
https://github.com/luvit/openssl/tree/master/asm/x64-win32-masm/aes
https://github.com/luvit/openssl/tree/master/asm/x64-elf-gas/aes
https://github.com/luvit/openssl/tree/master/asm/x64-macosx-gas/aes

Please, try to enable to compile the code for AES-NI in LibreSSL for the Windows 64bit target.
Thank you!

busterb:
Thanks for the note. There is no specific technical reason it's disabled, just a matter of no-one having done the work yet.

lars18th:
> Thanks for the note. There is no specific technical reason it's disabled, just a matter of no one having done the work yet.

Great! So, I hope someone will enable it. Then the Windows OpenSSH _native_ server will adquire AES-NI support.

Regards. :wink:

lars18th:
Hi,
Is anyone working on this?

busterb:
I had a notion to work on it at the hackathon this week. Patches welcome of course.

feliwir:
Any progress regarding this topic? I read the following note in the CMakeLists.txt:
```
# XXX Disabled for now, CMake's MASM support seems to either never
# build supply ASM or build it with the C compiler in a mode where it
# does not parse correctly. It might be easier to get NASM support working.
```

feliwir:
I've figured why the problem described above occurs. According to this document:
https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/languages/Assembler
The extensions for NASM support in CMake must be *.asm. Insead we call `set_property(SOURCE ${ASM_X86_64_MASM_SRC} PROPERTY LANGUAGE C)` which makes it not use the MASM assembler (check for Item Type). When changing the extension to *.asm and removing that property it works as expected.

ValZapod:
https://github.com/libressl-portable/portable/pull/526 fixes this...

PR #525, opened 2019-05-15
[WIP] Enable masm compilation under windows

feliwir:
Adresses some issues with the cmake configuration. *.masm files are now compiled with masm as part of the build process. Adresses #474 
TODO:
- [x] Remove include directives from masm assembly files
- [ ] Validate AES-NI support is working


CMakeLists.txt
```diff
@@ -289,7 +289,7 @@ if(ENABLE_ASM)
 		endif()
 	elseif(APPLE AND "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
 		set(HOST_ASM_MACOSX_X86_64 true)
-	elseif(MSVC AND "${CMAKE_GENERATOR}" MATCHES "Win64" AND FALSE)
+	elseif(MSVC AND "${CMAKE_SYSTEM_PROCESSOR}" MATCHES "AMD64")
```
feliwir: The old fashion of checking for Win64 doesn't work anymore with VS2019

crypto/CMakeLists.txt
```diff
@@ -147,7 +147,7 @@ if(HOST_ASM_MASM_X86_64)
 	add_definitions(-DWHIRLPOOL_ASM)
 	add_definitions(-DOPENSSL_CPUID_OBJ)
 	set(CRYPTO_SRC ${CRYPTO_SRC} ${ASM_X86_64_MASM_SRC})
-	set_property(SOURCE ${ASM_X86_64_MASM_SRC} PROPERTY LANGUAGE C)
+	set_property(SOURCE ${ASM_X86_64_MASM_SRC} PROPERTY LANGUAGE ASM_MASM)
```
feliwir: Files are getting picked up correctly by VS when setting the language to ASM_NASM

busterb:
To actually build the .asm files under `masm`, we should also pipe them through `cpp` in update.sh as part of the generation process, so there are no `#include` directives in the final product. Let me see if I can whip up a diff.

busterb:
Try adding something like this and run 'update.sh' again:
```
--- a/update.sh
+++ b/update.sh
@@ -167,7 +167,11 @@ gen_asm_stdout() {
        .section .note.GNU-stack,"",%progbits
        #endif
        EOF
-       $MV $3.tmp $3
+       if [ $1 = "masm" ]; then
+               cpp -I./crypto $3.tmp > $3
+       else
+               $MV $3.tmp $3
+       fi
 }
 gen_asm() {
        CC=true perl $asm_src/$2 $1 $3.tmp
@@ -176,7 +180,11 @@ gen_asm() {
        .section .note.GNU-stack,"",%progbits
        #endif
        EOF
-       $MV $3.tmp $3
+       if [ $1 = "masm" ]; then
+               cpp -I./crypto $3.tmp > $3
+       else
+               $MV $3.tmp $3
+       fi
 }
```

That'll preprocess the masm-targeted assembly, though I don't have a machine to test at the moment to verify the output is parsable by it other than validating the `#include` directives are gone and the defines are expanded into constants.

feliwir:
@busterb 
How am i supposed to run that on windows actually?😅 I’ve only used release source trees. When running with mingw i’ve failed.

busterb:
Locally I just run it on Linux, OpenBSD or MacOS and my Windows targets just build from a share. That's the fastest way without having to sort out how to get a posix shell on Windows.

The script also runs fine from cygwin assuming you have installed the prerequisites mentioned in README.md: automake, autoconf, git, libtool, perl. That's what I've used locally in the past, and the environment under which the `dist-win.sh` script was run. I suppose with the above diff you'll also need to at least add the base gcc package to cygwin too to ensure that `cpp` is available.

The `appveyor.yml` file shows a few of the msys64 pacman commands that are used to configure the environment with that tool as well, if that's your preference.

busterb:
Maybe using `WSL` is another choice, though I'm not sure how well file sharing is between it and the base OS until `WSL2` is available.

Anyway, I'm doing a quick test build to get appveyor back on track and to see if the `cpp` update will work on that environment as well. https://github.com/libressl-portable/portable/pull/526

busterb:
Tests in #526 look alright (there was a github timeout, but that's not a big problem). I'll merge your changes in too and we should get some initial results.

busterb:
Nope, needs a little more work (at least when generating on Windows): https://ci.appveyor.com/project/busterb/portable/builds/24616703/job/0pc9rnfdxol9ockc

feliwir:
That’s strange, i’ll check it out when i’m home

feliwir:
So the issue is the OPENSSLDIR variable. All the flags that were set my `add_definition` are passed to MASM aswell, but it seems like the whitespace inside the OPENSSLDIR is causing a failure. E.g. i tried to open 
C:/Program **Files/LibreSSL/etc/ssl**, which is the part after the whitespace. 
Since building always somewhere else is not an option i recommend setting the OPENSSLDIR property on a source level base, instead of defining it globally

feliwir:
Any idea on how to process @busterb ?

busterb:
I just cherry-picked @kinichiro 's fix into #526 . How about we focus on that PR for resolving the remaining masm build issues. Letting it run against travis + appveyor now, and will see how it fares in a bit.

busterb:
Closing this in favor of #526

PR #526, opened 2019-05-17
Windows build fixes

busterb:
This pulls in changes from #525 moves forward additional build fixes.


kinichiro:
It seems that '#'(for comments), '|', '~' and '<<' are treated as invalid with MASM.
After replacing these operators, compilation was completed.
See this commit. https://github.com/kinichiro/portable/commit/75d1b7223f79134d126b9cad6e45f7511c93925b

I'm not sure this sed replacing is the best way to solve this ...

commit above modifies below, too.
- To judge adding `-DOPENSSL_NO_ASM` or not, using `CMAKE_GENERATOR` instead `CMAKE_SYSTEM_PROCESSOR`.
- To avoid "cannot open file : Files/LibreSSL/etc/ssl", added `-DCMAKE_INSTALL_PREFIX`.

kinichiro:
Sorry, last commit breaks travis build, and here is fix.
https://github.com/kinichiro/portable/commit/92b366d81deed9cec8d717ad58e7331781f8f440

feliwir:
Do we not want to fix masm working under pathes with whitespaces?

busterb:
Sure @feliwir I saw your note on the other PR. Got a diff?

feliwir:
@busterb no, i think it would be a bigger CMake change, but maybe there's a better solution. We just need to get rid of the `add_definition` here: https://github.com/libressl-portable/portable/blob/master/crypto/CMakeLists.txt#L928 

I can try luck on the weekend on this issue maybe.

busterb:
I think the solution may have been as simple as 83012fe. Giving it a test now, but local results are positive.

busterb:
Hah, incremental compile hid a different issue. Hmm!

busterb:
I think I'm going to live with this for now.

feliwir:
So we merge it like this?

busterb:
I think the final revision looks reasonable. Don't want to default to something directly off of c:\, since anyone can create that on Windows if it doesn't already exist, and thus configure arbitrary certificate trusts and such.

feliwir:
I agree. Will there be a release with this feature in the near feature? Would be quite helpful for me, since AES-NI support has been something we need on windows.

busterb:
In the next dev release, sure! I hesitate to enable it in the 2.8.x tree though.

One note, when we enabled similar optimizations on Linux, in many systems the default linker would assume that the stack needed to be executable unless we overrode the behavior. Is there a similar override/check we should make with Microsoft's assembler to avoid inadvertently making the stack executable with libcrypto?

feliwir:
Not sure about that, sorry :-/

busterb:
Should be able to take a look today to verify.

busterb:
Looks fine to me!

(waiting for Cunningham's law to kick in!)

kinichiro:
It is good to me, and I have no idea about more tests on Windows MASM at
this time.

(Anything we couldn't find out will be posted after cutting out the
Release, according to the Murphy's law :)

>

PR #541, opened 2019-07-14
Enable speed on win32

kinichiro:
This PR enables openssl(1) speed command on WIN32 platform.
I referenced to OpenSSL implementation, but I put thread functions in
speed_alarm()/apps_win.c rather than Time_F()/speed.c.
We can put WIN32 calls in apps_win.c rather than speed.c,
and I think this way is near to original non-WIN32 implementation structure.
To remove timer thread, I added speed_alarm_free() and called it in
print_result().


busterb:
Seems reasonable, thanks @kinichiro

Issue #542, opened 2019-07-18
Win64: tls_load_config_file and text mode


ulfworsoe:
On Windows using Visual Studio `tls_config_set_ca_file` fails if newlines in the file are `\r\n` instead of `\n`. In reality this is a problem in `tls_config_load_file`. 

The problem seems to be that `stat()` gives the correct number of bytes in the file, but the file is opened in text mode and `read()` counts only one byte per line break. This causes the function to fail.

The solution would probably be to force binary mode in Windows.

Observed with: LibreSSL built with Visual Studio 2015 for AMD64 with static runtime (`/MT`).

kinichiro:
On Windows platform, open() with `O_BINARY` mode seems solve that problem.
Does this patch works good for you ?

```diff
--- tls_config.c.orig   Fri Jul 19 23:35:05 2019
+++ tls_config.c        Fri Jul 19 23:35:33 2019
@@ -52,7 +52,7 @@ tls_config_load_file(struct tls_error *error, const ch
        *buf = NULL;
        *len = 0;

-       if ((fd = open(filename, O_RDONLY)) == -1) {
+       if ((fd = open(filename, O_RDONLY|O_BINARY)) == -1) {
                tls_error_set(error, "failed to open %s file '%s'",
                    filetype, filename);
                goto err;
```

And same issue affects these 5 functions.
- CA : tls_config_set_ca_file()
- CRL : tls_config_set_crl_file()
- certificate : tls_keypair_set_cert_file()
- key : tls_keypair_set_key_file()
- ocsp : tls_keypair_set_ocsp_staple_file()

ulfworsoe:
That works for me.

kinichiro:
@ulfworsoe can you close this issue ?
PR #543 was merged and I think this issue was resolved.

ulfworsoe:
Thank you! I think this fixes it, so I'll close.

PR #543, opened 2019-07-26
Force binary mode for open() on WIN32

kinichiro:
This PR is for #542 and just adds O_BINARY for open() on WIN32 platform.
Should we solve this by other way rather than by patch as this PR ?
Thoughts ?


busterb:
Is there any reason not to just convert all `open `calls via the posix intercept layer we have to be unconditionally binary-only for consistency? I think we already do this with our `posix_open` intercept in posix_win.c in libcrypto.

kinichiro:
Yes, we can use posix_open().
I thought posix_win.c should adapt only for winsock operations, but it was my misunderstood.
I added another commit that removes 3 patches that kept from using posix_*() functions.
I also modified posix_win.c to handle WSANOTINITIALISED that happens when WSAStartup() is not called.
I noticed that keypairtest fails since it got WSANOTINITIALISED in posix_read().

ok?

PR #549, opened 2019-09-20
Check for existing uninstall target before creating.

mattparks:
Prevents LibreSSL from generating it's own uninstall target if one already exists. In big CMake projects there is usually one already defined and will cause a error similar to this:
```
1> [CMake] CMake Error at Build/x64-Debug/_deps/libressl-src/source/CMakeLists.txt:259 (add_custom_target):
1> [CMake]   add_custom_target cannot create target "uninstall" because another target
1> [CMake]   with the same name already exists.  The existing target is a custom target
1> [CMake]   created in source directory
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -357,10 +357,12 @@ if(NOT MSVC)
 		DESTINATION ${CMAKE_INSTALL_LIBDIR})
 endif()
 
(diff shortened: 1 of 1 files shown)


busterb:
Makes sense, thanks.

PR #550, opened 2019-10-08
Add OSS-Fuzz fuzzing badge

inferno-chromium:


inferno-chromium:
@busterb - can you please take a look, would love if you can help to spread word about OSS-Fuzz.

PR #557, opened 2019-11-16
Override the default path of cert.pem in libtls

kinichiro:
This PR is for issue https://github.com/libressl-portable/portable/issues/556 .
Refer to the commit https://github.com/libressl-portable/openbsd/commit/461e241c9b6c0a199a3f128d34ec82a7f2cc1cab , too.


kinichiro:
regress tlsexttest fails on linux(gcc), arm32(gcc) and arm64(gcc), but this failure appears to be caused by other reason.

PR #574, opened 2020-03-25
crypto: do not include gost.h into noinst_HEADERS

lumag:
gost/gost.h will go into include/openssl/, no need to duplicate it into crypto/gost directory

Signed-by: Dmitry Baryshkov <dbaryshkov@gmail.com>


lumag:
Travis has failed Android builds without actually going into build stage. Should I fix my Travis setup somehow?

kinichiro:
Travis-CI build failure was another issue and I posted PR for that.
I think your fix is good.

PR #576, opened 2020-03-31
crypto: regenerate crypto_portable.sym

lumag:
Make crypto_portable.sym depend on crypto.sym and libcrypto.so on
crypto_portable.sym to rebuild library if one of symbol files changes.

Signed-off-by: Dmitry Baryshkov <dbaryshkov@gmail.com>


busterb:
Nice change, thanks

PR #578, opened 2020-04-15
Add support for use of static MSVC runtimes

OJ:
In certain contexts LibreSSL needs to be built with `/MT` instead of `/MD` for inclusion in other projects. These changes allow for the command-line option `USE_STATIC_MSVC_RUNTIMES` to be set to `ON` if the use wishes to generate projects that will build with static runtimes.

This feature requires CMAKE version 3.15+, hence the minimum required version has changed as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.0)
+cmake_minimum_required (VERSION 3.15)
 include(CheckFunctionExists)
 include(CheckSymbolExists)
 include(CheckLibraryExists)
@@ -9,6 +9,7 @@ set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}" ${CMAKE_MODULE_PATH})
 include(cmake_export_symbol)
 include(GNUInstallDirs)
 
+cmake_policy(SET CMP0091 NEW)
 project (LibreSSL C ASM)
 
 enable_testing()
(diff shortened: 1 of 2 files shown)


busterb:
Howdy and thanks @OJ . Looks like this change has some issue on 64-bit builds with MSVC when inline assembly is checked for during the configuration phase. Might need to tweak so that these flags aren't passed to MASM.

busterb:
Looks like this just needs a bump at least to CMake 3.15.7 as a minimum, the bug was in CMake itself: https://gitlab.kitware.com/cmake/cmake/issues/20236

OJ:
Oh nice! I shall fix. THanks.

kinichiro:
Original issue appeared to be reported with 'cmake 3.16.0.',
so Milestone 3.15.7 seems strange to me.
https://gitlab.kitware.com/cmake/cmake/issues/20236

Related merge request was targetting 3.16.4, here.
I think this issue would be fixed in 3.16.4.
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4257

And cmake_minimum_required can't force the version of cmake.
It just quit to process if the running cmake version is lower than that.

Currently, Appveyor build is done with "Visual Studio 2015".
And pre-installed cmake version for VS2015 is 3.16.2, as described here.
https://www.appveyor.com/docs/windows-images-software/

I guess updating to VS2019 might succeed to build in Appveyor.

kinichiro:
For Appveyor fails, `-A x64` will work.
https://cmake.org/cmake/help/git-stage/generator/Visual%20Studio%2016%202019.html


For Travis-ci fails, how about adding `if(MSVC)` around `cmake_policy(SET CMP0091 NEW)`.

```
CMake Error at CMakeLists.txt:16 (cmake_policy):
Policy "CMP0091" is not known to this version of CMake.
```

CMP0091 appears to be introduced with cmake 3.15.
https://cmake.org/cmake/help/git-stage/policy/CMP0091.html

busterb:
Hah, of course they would switch from 'Win64' to 'x64', but make 32-builds 'Win32' :)

busterb:
Woo, that worked, merging!

OJ:
Great, thanks everyone.

PR #579, opened 2020-04-15
avoid exporting a sleep() symbol from libcrypto

busterb:
Since it seems only MSVC lacks sleep(), and it's only used by apps, lets special-case that and make it available as a static inline function instead. This fixes linkage issues where the target app also defines a sleep() compatibility function.


busterb:
It might be better to add a more libressl-specific namespace in front of all of the posix_ functions too. So far haven't noticed collisions with other apps, so no real rush.

kinichiro:
looks ok to me.

PR #585, opened 2020-05-01
switch enable-tests to enable_tests

vtorri:


kinichiro:
I had confirmed that `./configure --disable-tests` disabled building and running regression tests.

kinichiro:
Sorry, misunderstood. This defaults to disable testing.

configure.ac
```diff
@@ -66,6 +66,10 @@ AC_ARG_ENABLE([extratests],
 	AS_HELP_STRING([--enable-extratests], [Enable extra tests that may be unreliable on some platforms]))
 AM_CONDITIONAL([ENABLE_EXTRATESTS], [test "x$enable_extratests" = xyes])
 
+AC_ARG_ENABLE([tests],
+	AS_HELP_STRING([--enable-tests], [Enable tests]))
```
kinichiro, line 70: I think enable_tests should be default to yes.

```
AC_ARG_ENABLE([tests],
        AS_HELP_STRING([--enable-tests], [Enable tests]),
        [],
        [enable_tests=yes])
```

vtorri:
i'm right now improving it, thanks

configure.ac
vtorri: it should be good now

configure.ac
```diff
@@ -66,6 +66,16 @@ AC_ARG_ENABLE([extratests],
 	AS_HELP_STRING([--enable-extratests], [Enable extra tests that may be unreliable on some platforms]))
 AM_CONDITIONAL([ENABLE_EXTRATESTS], [test "x$enable_extratests" = xyes])
 
+AC_ARG_ENABLE([tests],
+       [AS_HELP_STRING([--disable-tests], [Enable tests (default: yes)])],
```
kinichiro, line 70: If using "--disable-tests", help message would be "Disable tests" ?
IMHO, default value description is not needed like "--disable-asm".

configure.ac
```diff
@@ -66,6 +66,16 @@ AC_ARG_ENABLE([extratests],
...
+       [if test "x$enable-tests" = "xyes"; then
```
kinichiro, line 71: s/enable-tests/enable_tests/
This causes "--enable-tests", "--enable-tests=yes", "--enable-tests=no" gives disabling tests.

kinichiro: And I wonder if we could do this only with "enable_tests" without "have_tests" to keep simple.

configure.ac
```diff
@@ -66,6 +66,15 @@ AC_ARG_ENABLE([extratests],
...
 
+AC_ARG_ENABLE([tests],
+       [AS_HELP_STRING([--disable-tests], [Disable tests @<:@default=yes@:>@])],
+       [
+        if ! test "x${enable_tests}" = "xyes"; then
+		enable_tests="no"
+	fi],
+       [have_tests="yes"])
+AM_CONDITIONAL([ENABLE_TESTS], [test "x$have_tests" = xyes])
+
```
kinichiro, line 77: This does not work properly when using "--enable-tests" and "--enable-tests=yes", and it disables tests.
Help string shows only "--disable-tests" though, I think "--enable-tests" also has to work correctly.
This uses 2 variables "enable_tests" and "have_tests", but I believe we can handle this situation by "enable_tests" only.

Changes below could handle all of "--disable-tests", "--enable-tests", "--enable-tests=yes", "--enable-tests=no" properly.

```
AC_ARG_ENABLE([tests],
       AS_HELP_STRING([--disable-tests], [Disable tests]),
       [],
       [enable_tests=yes])
AM_CONDITIONAL([ENABLE_TESTS], [test "x$enable_tests" = xyes])
```

Can you try this ?

vtorri: for the help, why adding help for --enable-tests if tests are enabled by default ? Packagers are not stupid

kinichiro: I did not mean that we have to add help string for "--enable-tests".
I wanted to say "--enable-tests" option has to work correctly as same as "--disable-tests".

And Your last commit appears to work right.

Issue #602, opened 2020-07-12
build system and __STRICT_ALIGNMENT


q66:
I've been poking around LibreSSL lately wrt importing assembly support for ppc(64(le)) and aarch64, and ran into some curious things:

1) there is the `__STRICT_ALIGNMENT` macro which seems to control whether the platform does or doesn't deal with supporting unaligned memory access; there is a similar macro in OpenSSL, called `STRICT_ALIGNMENT`
2) however, OpenSSL by default forces it on, except on platforms where it seems to be safe: https://github.com/openssl/openssl/blob/33388b44b67145af2181b1e9528c381c8ea0d1b6/include/crypto/modes.h#L30 Currently, these platforms are the x86 family, aarch64 and s390(x). ARMv7 used to be allowed, but got disabled at one point https://github.com/openssl/openssl/commit/29a546720b6d8ea8b3207f500e3ed9d2834b09f2 - that's why there are the `BSWAP4` and `BSWAP8` assembly macros, despite seemingly never being used.
3) LibreSSL drops the code that does this, which means `STRICT_ALIGNMENT` is always not defined by default.
4) The build system then conditionally defines it: https://github.com/libressl-portable/portable/blob/master/configure.ac#L78

As I see it, there are some issues with this.

1) This completely missed all those other platforms for which OpenSSL defines the macro but LibreSSL then doesn't
2) That check doesn't make much sense to me - as it seems to intend to work, it will pretty much pass for x86 family and 32-bit ARM and fail for everything else. That's because `BSWAP4` is defined when on ARM or x86 and when `__STRICT_ALIGNMENT` is not defined, but not under any other circumstances.
3) Another piece of code then implies it at least for SPARC, but nothing else
4) Then there is this line: `AS_IF([test "x$BSWAP4" = "xyes" -a "$host_cpu" = "arm" ],,CPPFLAGS="$CPPFLAGS -D__STRICT_ALIGNMENT")` This basically means that `__STRICT_ALIGNMENT` is always defined unless on 32-bit ARM. That was added by 8f32297dcedbb883b29f7a9e102dd34c7b69af6c and the previous behavior was different. Now, `__STRICT_ALIGNMENT` will be defined on x86, while previously it was not.
[...]

PR #604, opened 2020-07-21
Get __STRICT_ALIGNMENT from machine/endian.h

kinichiro:
Proposal for #602

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -76,7 +76,6 @@ AC_ARG_ENABLE([tests],
 AM_CONDITIONAL([ENABLE_TESTS], [test "x$enable_tests" = xyes])
 
 AS_CASE([$host_cpu],
-	[*sparc*], [CPPFLAGS="$CPPFLAGS -D__STRICT_ALIGNMENT"],
 	[*arm*], [host_cpu=arm],
 	[*amd64*], [host_cpu=x86_64 HOSTARCH=intel],
 	[i?86], [HOSTARCH=intel],
--- a/configure.ac
+++ b/configure.ac
@@ -75,18 +75,6 @@ AC_ARG_ENABLE([tests],
        [enable_tests="yes"])
 AM_CONDITIONAL([ENABLE_TESTS], [test "x$enable_tests" = xyes])
 
-# Add CPU-specific alignment flags
-old_cflags=$CFLAGS
-CFLAGS="$CFLAGS -I$srcdir/include"
-AC_MSG_CHECKING([if BSWAP4 builds without __STRICT_ALIGNMENT])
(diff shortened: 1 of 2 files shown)


q66:
This looks fine to me in principle, there are still some concerns though:

1) according to OpenSSL, this should also be covered for `__s390__` and `__s390x__`
2) they seem to imply, in https://github.com/openssl/openssl/commit/29a546720b6d8ea8b3207f500e3ed9d2834b09f2, that it's not actually safe to enable for ARMv7. However, defining `__STRICT_ALIGNMENT` for ARMv7 in LibreSSL would result in the assembly implementation of `ghash` getting disabled. OpenSSL doesn't have a restriction like that in place for `ghash` assembly (they have a much newer version though), and they define the macro for all 32-bit ARM. According to your mentioned OpenBSD commit, LibreSSL upstream also had the macro enabled for all 32-bit ARM.

Also, if you add this, you might want to drop the special case for SPARC from the `configure.ac` script, as it becomes redundant.

kinichiro:
Thanks for seeing this PR and for your comments.
I added one commit.

1. I added arch macro for s390/s390x
2. I think it is the point that give difficulties about ARM with alignment issue.
   v6 and above has some kind of unaligned access support, but there seems to be some conditions ...
   This architecture checks in <machine/endian.h> should be changed according to the progress of ARM.
   Do you have any concrete recommendations for this check, now ?

I also removed redundant sparc check from configure.ac.

q66:
I guess I'd leave it as it is in your PR right now, to match the original intent of the previous code (as `BSWAP4` is defined for ARMv7 without `__STRICT_ALIGNMENT`) and it can perhaps be changed later if the assembly code gets updated

kinichiro:
@q66 Thanks for your feedback.

q66:
Found another thing: `crypto/gost/gost2814789.c` should `#include <machine/endian.h>`, otherwise it will not pick up the macro - I inspected other files as well and this is the only instance where the include is missing

q66:
There is one more thing that came to mind. The `__STRICT_ALIGNMENT` macro is used by some of the ARM `.S` files. That means if we move the macro definition into `endian.h`, it will not get picked up

kinichiro:
I think gost2814789.c should include <machine/endian.h>.

Currently \*armv4\*.S files seem to get \_\_STRICT_ALIGNMENT from arm_arch.h,
and it defines \_\_STRICT_ALIGNMENT when \_\_OpenBSD__ is defined.

busterb:
With the latest gost fixes, is this ready to merge @kinichiro ?

kinichiro:
@busterb I didn't get ok for gost diff yet. I will try again.

kinichiro:
@busterb I had committed gost fix. I think this is ready to merge.

PR #610, opened 2020-08-20
Configure libtls and nc(1) to statically link to libcrypto/ssl

busterb:
An issue that Reyk Floeter noted while building a Debian package for LibreSSL is that installing libtls along with OpenSSL causes linker issues since it will often pick up the wrong libcrypto/libssl. This change makes libtls statically link the object files it needs rather than relying on the shared libraries, effectively making libtls self-contained and able to be packaged independently.

This should make it possible for other projects that also use libtls to be able to package support without requiring the target OS to ship libcrypto / libssl from LibreSSL.

See
https://salsa.debian.org/reyk-guest/libressl/-/commit/678278df55ce866f2f363998ca690442fa786c66


CameronNemo:
What would be really nice is if the libressl symbols were amalgated so as to not conflict with openssl if both are linked in the same program.

This is a great start though, and I cannot find any program which links both libtls and libcrypto or libssl.

busterb:
Agree @kinichiro I don't really see a way to do this portable without linking all of the object files directly. Sharing the object file list might require putting some of the common definitions into Makefile.am.common instead, since as things are now, those definitions aren't shareable I think.

kinichiro:
@busterb I was thinking about these kind of things https://github.com/kinichiro/portable/commit/497662407a67abcd69edf4cc770a09832284f297 .
It generates libtool object files list variable from crypto and ssl, then link them directly to libtls.
This might be an irregular way to work with autotools, though.
This succeeds on OpenBSD, but other platforms `make` doesn't have `ifmake target(...)` functionality and fail.

kinichiro:
Now it succeeds to build with this commit https://github.com/kinichiro/portable/commit/3e86059f6361df9c0b9e2cffdb032b5f8cf82f81
I found that nc requires to link with libcrypto since getentropy_linux.c uses SHA512_Update in it.

busterb:
I pulled your changes in @kinichiro and this looks to work, thanks!

PR #620, opened 2020-09-14
Add x509_issuer_cache.c, regress x509attribute.c

kinichiro:
This PR adds x509_issuer_cache.c and x509_issuer_cache.h, and regress x509attribute.c.

x509_issuer_cache.h includes `sys/queue.h` `sys/tree.h`, and those require `sys/_null.h`.
But Linux or Windows platform seems not to have these header files.
I thought these should be provided as include/compat/ files.

With this PR, I've just added these three header files from OpenBSD `/usr/src/sys/sys/`.
Another way might be importing `/usr/src/sys/sys/` into `libressl-portable/openbsd` repository and copying those 3 headers by `update.sh` as doing for other source codes.

I thought this PR way would be good and simple, though, I would like to hear which is better.


busterb:
Either way is fine. The tree.h and queue.h APIs are stable enough anyway.

PR #621, opened 2020-09-15
Add x509_verify.c x509_verify.h

kinichiro:
This PR is
- Add x509_verify.c x509_verify.h
- Restrict cipherstest with static build only in CMake, since it changes to use internal function
- Add regress x509_info.c
- Update .gitignore for some files in tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ tests/bnaddsub*
 tests/bn_rand_interval*
 tests/bn_to_string*
 tests/cipher*
+tests/constraints*
 tests/explicit_bzero*
 tests/freenull*
 tests/gost2814789t*
@@ -77,6 +78,9 @@ tests/*.pem
 tests/testssl
 tests/*.txt
 tests/compat/*.c
+tests/verify*
+tests/x509_info*
+tests/x509attribute*
 tests/x509name*
 !tests/optionstest.c
 !tests/*.test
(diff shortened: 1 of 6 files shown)


kinichiro:
Some regress getting fails

on Windows
- cipherstest (Failed)
- cmstest (SEGFAULT)
- ssltest (Failed)
- tlstest (Failed)

on arm32 and arm64
- cipherstest

OSX, Linux, Android seems fine.
Need investigation.

busterb:
for arm and aaarch64 cipherstest failures (I haven't looked at the Windows ones yet)

```
FAIL: test 0 - got cipher 0 with id 3001303, want 3001302
FAIL cipherstest (exit status: 1)
```

busterb:
For more context, it's failing in `cipher_set_test`

Removing the `goto failed`, we can see that the ciphers are being returned in a different order than expected by the test on arm/aarch64 (though this may be expected by the library):

```
FAIL: test 0 - got cipher 0 with id 3001303, want 3001302
FAIL: test 0 - got cipher 1 with id 3001302, want 3001303
FAIL: test 1 - got cipher 0 with id 3001303, want 3001302
FAIL: test 1 - got cipher 1 with id 3001302, want 3001303
```

busterb:
I think I'll land this PR for now, since it is correct on its own. The other issues should be addressed separately.

kinichiro:
Thanks! It helps to work for now.

PR #658, opened 2021-04-05
Add '--enable-libtls-only' configure option

busterb:
This adds an option for only installing libtls and its headers and manpages, skipping libcrypto and libssl. This is useful for taking advantage of the static-linked libtls that was added in LibreSSL portable 3.2.2, but when the baseline libcrypto/libssl on the target OS is OpenSSL. The new `--enable-libtls-only` option still builds libcrypto/libssl since they are needed internally, but does not install them or their dependencies. This moves some OS-dependent checks and installation of the certificate files to more correct places as well.

This is only implemented for autotools-based builds currently.


configure.ac
```diff
@@ -132,18 +129,14 @@ AC_CONFIG_FILES([
...
-        echo " ** mingw-w64 toolchain. Refer to README.windows for details."
-    fi
-fi
+AC_ARG_ENABLE([nc],
+	AS_HELP_STRING([--enable-nc], [Enable installing TLS-enabled nc(1)]))
+AM_CONDITIONAL([ENABLE_NC], [test "x$enable_nc" = xyes])
+AM_CONDITIONAL([BUILD_NC],  [test x$BUILD_NC = xyes -o "x$enable_nc" = xyes])
+
+AC_ARG_ENABLE([libtls-only],
+	AS_HELP_STRING([--enable-libtls-only], [Enable installing libtls only(1)]))
```
kinichiro, line 138: I think this "(1)" should be removed, and just "Enable installing libtls only" is fine.

kinichiro:
I had tested this PR on my local environment and works fine with option "--enable-libtls-only".
I also checked Windows build on my local PC and succeeded, too.
I think it is good to have this option.

busterb:
Thanks @kinichiro! I fixed the help text and squashed the commits, one more CI test and I'll merge it.

PR #668, opened 2021-05-16
cmake: tweak for clang-cl

lemourin:
This change allows libressl to be built with clang-cl.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,7 +113,6 @@ if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
 endif()
 
 if(WIN32)
-	add_definitions(-Drestrict)
 	add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 	add_definitions(-D_CRT_DEPRECATED_NO_WARNINGS)
 	add_definitions(-D_REENTRANT -D_POSIX_THREAD_SAFE_FUNCTIONS)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -113,7 +113,10 @@ if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
 endif()
 
 if(WIN32)
-	add_definitions(-Drestrict)
+	# Don't define 'restrict' keyword for clang-cl
+	if(NOT (MSVC AND CMAKE_C_COMPILER_ID MATCHES "Clang"))
```
busterb, line 117: Could we use 'if(CMAKE_C_COMPILER_ID MATCHES "MSVC")' like below so this is explicitly only hiding restrict from MSVC?

lemourin: Any reason why the `restrict` keyword has to be hidden at all? Latest MSVC seems to work without it being hidden. Do you need to support some ancient MSVC version which does not work with `restrict`?

busterb: Since it was added in the first commit, it's hard to say what purpose it served, other than it being from 7 years ago. The need possibly went away in 2018 with 3fb9e63b907186c70fb79e818d899ccb67b4b421 when Vista was made the minimum requirement and the builtin inet_ntop could be used.

If we can remove the line entirely now, even better; I'm not sure I see any current uses of the restrict keyword in the current code anyway.

lemourin: I removed the line. Thanks for the context. `grep -R restrict` in libressl source tree doesn't seem to show any actual usages.

busterb:
Looks good, thanks for the update!

PR #671, opened 2021-05-22
Use github actions for linux test

kinichiro:
This PR is experimental replacing travis-ci with github actions for linux test.

I think all the things that `scripts/travis` does could be written in `.github/workflows/linux_test.yml` as `steps:`.
But this PR just adds yaml file and uses `scripts/travis` almostly as is.
It might be better to rewrite this part in the future, if this is in.

Additionaly, I added another commit to force linking strtonum for macos cmake build, since I found testdsa and testrsa fail on macos 10.15.

One drawback with this github actions.
Now I set `continue-on-error to true`, and this causes workflow status as green even if some jobs fail.
If removing `continue-on-error`, one job failing brings whole workflow terminated on the way.
I wonder if which is better way for testing.

Any comments or thoughts ?


kinichiro:
We can see the result here
https://github.com/kinichiro/portable/actions/runs/865924244

busterb:
This looks great, thank you! If we have to choose, I would want failures to be evident, even if it means one failure aborts other tests. Most of the time when I see a failure because of an upstream change, it's due to something that would have failed multiple builds anyway. I can see wanting to go the other way when fixing the build itself, but most of the time, I think failing early and having a correct visual status is the way to go.

kinichiro:
Thanks for reviewing this.
I agree with you and added another commit to set continue-on-error to false.

busterb:
Great! A couple more questions, is it required by github actions to call the workflow linux_test.yml, or can we have multiple workflow files, maybe one per OS? What is required to get the status showing up as a check status in this PR? Can we change the README build status badge too?

kinichiro:
>is it required by github actions to call the workflow linux_test.yml, or can we have multiple workflow files, maybe one per OS?

I tried to split out the android test as another workflow file, and this resulted two different summary dashboard.
https://github.com/kinichiro/portable/actions/runs/874544394
https://github.com/kinichiro/portable/actions/runs/874544395
I feel like to see test results in one summary dashboard.
So I don't think https://github.com/libressl-portable/portable/pull/671/commits/31ef9ade60a19d369efc81528283d0c8735637cc is better.

>What is required to get the status showing up as a check status in this PR?

On my personal test repo, I did not set anything and just push .yml file.
"Settings" for repo has "Actions" menu, and "Actions permissions" is set to "Allow all actions".
This is the same as portable repo.
I suspect that is because currently portable master branch has no workflow yet.

>Can we change the README build status badge too?

Yes, but badge URI appears to have repo name, workflow name and workflow file name in it.
It should change to better name (i.e. workflow name = "build", workflow file = build.yml)
See below.
image

busterb:
Thanks! I'm going to land this and we can make any additional adjustments once there is an initial workflow in the tree.

kinichiro:
Yes, go ahead your additional adjustment, please.
Thanks.

PR #672, opened 2021-05-28
More github actions updates

busterb:
Following up on work from @kinichiro, this splits out a few more tests into logical actions so we can do more testing in parallel. It also renames the testing script and adds updated badges to the README.md.

Issue #649, opened 2021-02-16
Compiling Libressl 3.3.1 on Ubuntu for Windows x86 and x64 target failes 


kbsec:
Hey there!

I am trying to compile libressl shared libraries (libssl, libcrypto) on an ubuntu box for a windows target. 
my mingw version is 7.0.0-2

after running autogen.sh,  I ran
```
 CC=x86_64-w64-mingw32-gcc ./configure --host=x86_64-w64-mingw32
 make
 make check
```

And i get the following errors
```
CCLD     libcrypto.la
/usr/bin/x86_64-w64-mingw32-ld: asn1/.libs/libcrypto_la-p5_pbev2.o: in function `memcpy':
/usr/share/mingw-w64/include/string.h:202: undefined reference to `__memcpy_chk'
/usr/bin/x86_64-w64-mingw32-ld: bio/.libs/libcrypto_la-b_dump.o: in function `memset':
/usr/share/mingw-w64/include/string.h:208: undefined reference to `__memset_chk'
/usr/bin/x86_64-w64-mingw32-ld: des/.libs/libcrypto_la-enc_writ.o: in function `memcpy':
/usr/share/mingw-w64/include/string.h:202: undefined reference to `__memcpy_chk'
/usr/bin/x86_64-w64-mingw32-ld: dsa/.libs/libcrypto_la-dsa_gen.o: in function `memcpy':
/usr/share/mingw-w64/include/string.h:202: undefined reference to `__memcpy_chk'
/usr/bin/x86_64-w64-mingw32-ld: /usr/share/mingw-w64/include/string.h:202: undefined reference to `__memcpy_chk'
/usr/bin/x86_64-w64-mingw32-ld: /usr/share/mingw-w64/include/string.h:202: undefined reference to `__memcpy_chk'
/usr/bin/x86_64-w64-mingw32-ld: ec/.libs/libcrypto_la-eck_prn.o: in function `memset':
/usr/share/mingw-w64/include/string.h:208: undefined reference to `__memset_chk'
/usr/bin/x86_64-w64-mingw32-ld: /usr/share/mingw-w64/include/string.h:208: undefined reference to `__memset_chk'
/usr/bin/x86_64-w64-mingw32-ld: evp/.libs/libcrypto_la-e_aes_cbc_hmac_sha1.o: in function `memcpy':
/usr/share/mingw-w64/include/string.h:202: undefined reference to `__memcpy_chk'
[...]

PR #674, opened 2021-05-30
Fix mingw build on ubuntu 20.04

kinichiro:
I believe this PR would solve #649 and MinGW build failure on ubuntu 20.04 in #672 .
I saw "undefined reference to `__memcpy_chk'" error while building on ubuntu 20.04.
ubuntu 20.04 appears to install MinGW 7.0.0, and it requires to link libssp if we define "-D_FORTIFY_SOURCE=2".
I also disabled dtlstest with MinGW since it does not have poll.

### DIFF
--- a/m4/check-hardening-options.m4
+++ b/m4/check-hardening-options.m4
@@ -73,7 +73,9 @@ AC_DEFUN([CHECK_C_HARDENING_OPTIONS], [
 		CHECK_CFLAG([[-fno-strict-overflow]])
 
 		# _FORTIFY_SOURCE replaces builtin functions with safer versions.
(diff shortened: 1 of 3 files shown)


kinichiro:
First commit (https://github.com/libressl-portable/portable/pull/674/commits/420723ee1818e1b62300f845f7af6a1a0cba6398) could solve link error but I found it did not generate DLLs.
For the mingw cross build, removing "-D_FORTIFY_SOURCE=2" might be better than adding "-lssp".
With second commit (https://github.com/libressl-portable/portable/pull/674/commits/e07b5feba5103e662016158a2d77ca6bf1db12fb), I confirmed DLLs were generated.

PR #678, opened 2021-07-19
Make sure library is built with non-executable stack

tobhe:
The GNU toolchain on ELF platforms defaults to executable stacks in assembly files that are not explicitly annotated.
A single object with executable stack causes the whole lib to be marked as requiring executable stack.

We can work around this for now by passing `-Wa,--noexecstack` on ELF platforms if ASM is enabled.

The problem is documented in more detail at https://wiki.ubuntu.com/SecurityTeam/Roadmap/ExecutableStacks
Here's some more discussion from [BoringSSL](https://github.com/google/boringssl/commit/9c49713ba8280deb1a1fcd7d018045ce0850115f) and [OpenSSL]( https://github.com/openssl/openssl/issues/4575)


busterb:
Good catch, we handle this in autoconf, but didn't have a CMake check. The build errors here look incidental and unrelated to these changes. You should be able to rebase on master soon to validate that this builds on all of the test systems.

CMakeLists.txt
```diff
@@ -300,6 +300,7 @@ if(ENABLE_ASM)
 		elseif(CMAKE_SYSTEM_NAME STREQUAL "SunOS" AND "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "i386")
 			set(HOST_ASM_ELF_X86_64 true)
 		endif()
+		set(CMAKE_ASM_FLAGS "${ASM_FLAGS} -Wa,--noexecstack")
```
busterb, line 303: Defining HAVE_GNU_STACK would also work here. Should we go with belt-and-suspenders, or even modify autoconf to prefer -Wa,--noexecstack?

tobhe: Oh I totally missed the autoconf check. We can just reuse HAVE_GNU_STACK in cmake to reduce the diff.

busterb: Thanks @tobhe!

PR #696, opened 2021-10-24
Specify local include paths for each regressions individually

kinichiro:
For regressions, specify local include path for each source files instead of adding them to AM_CPPFLAGS or include_directories.
Now upstream has the same name local header files in ssl/ and tls/, and this brings some regressions build break.
This PR avoid that issue.

Also change the path of cert.pem for ocsp_test.
cert.pem was moved from apps/openssl/ to top directory.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,16 +1,7 @@
 include_directories(
 	.
-	../crypto/modes
-	../crypto/asn1
-	../crypto/x509
-	../ssl
-	../tls
-	../apps/openssl
-	../apps/openssl/compat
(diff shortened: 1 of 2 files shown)


busterb:
I think if we went this route later, I'd like to just put all of the ssl and tls tests in separate directories, closer to the openbsd tree. For now, I landed on a simpler fix in 

https://github.com/libressl-portable/portable/commit/04ab1061f305d11ba292668e7405dd73008a2c1e#diff-4461c617ceae0c7e0206622aacdf555aedd62eb129c2efb74c84fa1567bcbe0d

Thanks!

PR #701, opened 2021-11-23
Avoid accessing ASN1 item in regress x509req_ext.c on Windows platform

kinichiro:
Now regress x509req_ext fails on Windows shared build, since it can not access ASN1 item X509_EXTENSIONS_it in DLL.
I thought one solution would be a patch to disable calling ASN1_item_unpack.
It is better if we could disable it when shared build only, but I had no idea to do that.
Another solution would be off this test when Windows shared build.
Thoughts, comments or any other idea ?


busterb:
Would it be possible for the shared build to still always build static libraries, just not install them? That might remove some other special-casing we're doing in general.

kinichiro:
Currently I have no idea about that way, but would like to try to think of it.

kinichiro:
I would like to close since #703 was merged.

PR #703, opened 2021-11-27
Build static library for regression tests when shared build with CMake

kinichiro:
Another solution instead of #701 .
Build another static libraries when it build shared, and use them with regressions.
These static libraries are not installed.

commit e8ea73c247069e4e5592b31c0888a8ceb35f7773 in this PR temporarily avoid build break with crypto/ct and this should be removed after merging #702 .


kinichiro:
Currently this PR replaces all regressions linking library to static one.
I wonder if tests that are able to work with shared library should link with shared one ?

PR #705, opened 2021-12-04
Remove unneeded target_include_directories with cmake

kinichiro:
This could remove recurring of the same statement for include directories.
Instead of this removals, apps/* and tests  should have include path that
had been provided by INTERFACE_INCLUDE_DIRECTORIES of target libs and
internal static libs.


busterb:
Nice improvement, thanks.

PR #710, opened 2021-12-11
Add bytestring

kinichiro:
This PR adds crypto/bytestring/ to libcrypto.
Three bs source files are the same between crypto and ssl.
To avoid bs objects duplication in libtls, splits out bytestring related objects as internal library from libssl.
internal bs library should be linked to shared libssl, since bs related functions are not exported.
static libssl does not require it because those functions are provided from libcrypto.

commit a3a03842f68732cc08c0241c40cf99fd11a8a3a4 is same as #709


kinichiro:
I wonder if I should remove bs three objects from ssl/.libs/libssl.a after generating libssl.la.

PR #712, opened 2021-12-13
Ignore ar command error

kinichiro:
'ar d' command gets error on macos if specified object not exist.


busterb:
You know, I should just stop syncing upstream before pulling these fixes, there's always another place to update :smiling_face_with_tear:

kinichiro:
l always wait for syncing upstream to openbsd repo on github before posting
PR that would be required to build latest.

Is it better posting those PR before syncing upstream to github ?
In that case, PR ci build would fail, though.
Chicken or egg :)

PR #713, opened 2021-12-15
archive macos test logs to determine failure in DTLS tests

busterb:
Currently these seem to only fail in CI for me, so archive the logs to see what is happening.


busterb:
Hmm, flaky test passes here.

PR #717, opened 2021-12-27
enable scheduled Coverity scan

chipitsine:


chipitsine:
please note that Coverity scan requires proper COVERITY_SCAN_TOKEN specified in repo secrets

busterb:
Added the secret and doing a test run. The email might need to be busterb@gmail.com but we'll see if the token is sufficient.

chipitsine:
pull requests cannot access secret tokens

chipitsine:
sure, feel free to change email if needed

PR #737, opened 2022-02-27
update autotool macro deprecations

busterb:
fix latest round of autotool complaints


kinichiro:
This looks sane to me.
linux_ci_asan failure (FAIL explicit_bzero) would be solved by another PR.

PR #741, opened 2022-03-26
CI: switch ASAN builds to clang

chipitsine:


chipitsine:
gcc9 does not supress "explicit bzero" test.
clang or newer gcc do supress it

busterb:
Sounds good, thank you.

PR #765, opened 2022-06-14
tests: Fix the build with --disable-static

orbea:
NOTE: This PR depends on PR https://github.com/libressl-portable/openbsd/pull/132 being merged first.

The tests require `--enable-static` because `-export-symbols` is set in `libtls_la_LDFLAGS` which hides many of the internal symbols for the shared library.

This can be avoided by compiling most of the code into a convenience library which the tests and libtls can both link against. The installed shared library will still use `-export-symbols` as expected without conflicting with the tests that require access to internal symbols.

Fixes https://github.com/libressl-portable/portable/issues/754


orbea:
Can this PR please be reviewed?

busterb:
Hey @orbea when you get a chance check out #806 as a possibly simpler fix to this.

busterb:
Merged #806, thanks for the help.

PR #776, opened 2022-08-01
Fix XCode static lib generation

guusw:
Building for iOS using CMake's XCode generator causes issues when linking against the ssl/crypto targets because they don't contain source files directly.

It seems like XCode as an optimization does not output libssl.a and libcrypto.a causing library not found linked errors

This PR adds dummy source files so the libraries do get generated in this case


busterb:
Hi, I made a few small changes to this approach in #807, testing now.

busterb:
Closing this one out, thanks for the PR @guusw

PR #789, opened 2022-09-12
appveyor: store entire build folder in artifacts

chipitsine:


busterb:
This seems like a reasonable change, thanks.

PR #800, opened 2022-11-03
update config.sub and config.guess for M1 Mac

catatsuy:
config.sub and config.guess on LibreSSL are very old. To build LibreSSL on M1 Mac, we need to update config.sub.


busterb:
Thanks for the update @catatsuy

PR #801, opened 2022-11-08
Update checkout action to current version.

daztucker:
checkout v2 uses node.js v12 which is has been recently deprecated by github so switch to using the one on the main branch.  Workflows currently show the current warning:

Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/checkout

### DIFF
--- a/.github/workflows/android_test.yml
+++ b/.github/workflows/android_test.yml
@@ -12,7 +12,7 @@ jobs:
       MAX_NAL: 28
     name: Android 8.x 9
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@main
     - name: Run CI script
       run: ./scripts/test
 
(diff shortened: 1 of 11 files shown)


busterb:
Thanks! Looks like incidental patch fallout.

daztucker:
> Thanks! Looks like incidental patch fallout.

The current test failures are unrelated to this PR as it's also happening on master.

I came here to check this after seeing a failure on openssh building against libressl master:
https://github.com/openssh/openssh-portable/actions/runs/3415658082/jobs/5685021239
then I noticed the nodejs warnings, which I'd just fixed in openssh, hence this PR.

botovq:
Thanks. This should all be fixed now. Could you rebase your PR so we can see if this works before merging?

daztucker:
>  Could you rebase your PR so we can see if this works before merging?

Done.

botovq:
Merged, thanks.

PR #807, opened 2022-12-13
add direct source dependency to each library

busterb:
Fix library generation with some CMake generators by including a direct source file dependency for each library. This is a small variation on #776 that includes libtls

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -1009,7 +1009,7 @@ target_include_directories(crypto_obj
 	PUBLIC
 		../include)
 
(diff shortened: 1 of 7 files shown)


busterb:
appveyor failure was from a slow network, landing this

busterb:
It looks like the cluster fuzz target is failing because the Dockerfile it uses mixes files from the master branch + a pull request for building, so we end up with the wrong build.sh file being used for CI testing. This will work once landed to master: https://github.com/google/oss-fuzz/blob/master/projects/libressl/Dockerfile

PR #808, opened 2022-12-19
Internally link dependencies statically

busterb:
This removes some build complexity by linking bundled apps statically unconditionally. It also fixes a build error on some systems when there is trailing whitespace in the object list.


botovq:
I looked at this more closely and I think I like this a lot more than I thought initially. It solves a few problems we've had nicely. Let's roll this into 3.8.0 and see if anyone complains.

orbea:
This PR fixed implicit function declarations on musl systems when building openssh, thanks! https://github.com/gentoo/libressl/issues/549

Issue #814, opened 2023-02-08
Patch failure while building from source on master during ./autogen.sh


quapka:
I have tried to build `master` (`3c1e882d259e`) using the instructions from `README.md`. The very first step, i.e. running `./autogen.sh` fails:

```
❯ ./autogen.sh
pulling upstream openbsd source
Cloning into 'openbsd'...
remote: Enumerating objects: 144811, done.
remote: Counting objects: 100% (26496/26496), done.
remote: Compressing objects: 100% (13585/13585), done.
remote: Total 144811 (delta 12231), reused 23554 (delta 11745), pack-reused 118315
Receiving objects: 100% (144811/144811), 50.70 MiB | 10.45 MiB/s, done.
Resolving deltas: 100% (103980/103980), done.
Updating files: 100% (13680/13680), done.
Already on 'master'
Your branch is up to date with 'origin/master'.
Already up to date.
libcrypto version 50:1:0
libssl version 53:1:0
libtls version 26:1:0
LibreSSL version 3.7.1
copying libcrypto source
generating arm ASM source for elf
generating x86_64 ASM source for elf
generating x86_64 ASM source for macosx
generating x86_64 ASM source for masm
generating x86_64 ASM source for mingw64
copying libtls source
copying nc(1) source
copying ocspcheck(1) source
copying openssl(1) source
copying libssl source
copying tests
patching file tests/aeadtest.c
patching file include/openssl/bio.h
patching file crypto/bn/bn_div.c
Hunk #1 FAILED at 89.
1 out of 1 hunk FAILED -- saving rejects to file crypto/bn/bn_div.c.rej

❯ bat crypto/bn/bn_div.c.rej
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: crypto/bn/bn_div.c.rej
───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ --- openbsd/src/lib/libcrypto/bn/bn_div.c   2018-08-03 22:40:06.596555542 -0500
   2   │ +++ crypto/bn/bn_div.c  2019-01-20 19:26:37.891450640 -0600
   3   │ @@ -89,7 +89,7 @@
   4   │         q;                  \
   5   │     })
   6   │  #  define REMAINDER_IS_ALREADY_CALCULATED
   7   │ -#  elif defined(__x86_64)
[...]

PR #815, opened 2023-02-09
point OPENBSD_BRANCH to libressl-v3.7.0 for now

busterb:
re https://github.com/libressl/portable/issues/814#issuecomment-1423881715 temporarily point master to the 3.7.0  tag.


busterb:
Realizing it's too late for this, since master has moved on since 3.7.0

PR #843, opened 2023-04-17
link signertest to build

botovq:


botovq:
@busterb if you have time to take a look, it would be nice to have this available soon. op is working on this from the smtpd side.

PR #865, opened 2023-05-27
fix asm on x86_64 macos

busterb:
This should fix asm on macOS x86_64.

I don't believe there are any Apple Intel Macs that support CET anyway, as it seems they stopped at Generation 10.


botovq:
Is there no way to pass something like -Dendbr64='' via AFLAGS (or
similar) to the assembler?

PR #878, opened 2023-07-04
Add Windows builds to Github actions

busterb:
This can replace Appveyor, and has some benefits in that all the folks in this project can also see the results directly, and we can use this to push build artifacts to the releases.

Also tacks on a small tweak to macOS builds so you can tell them apart by architecture.


busterb:
This is off to a good start, even found a failing test in ... ssltest!

busterb:
Got all the variations working and rebased. One more test and this is ready to land.

PR #889, opened 2023-07-07
Remove hard-coded optimization level in CMake builds

### LINKED ISSUE #683: Incorrect compiler flags set in debug builds for MSVC and CMake
The following warning was observed when attempting to build a debug build with CMake/Ninja/MSVC:
```
cl : Command line warning D9025 : overriding '/O2' with '/Od'
```

I believe this line in CMakeLists.txt is causing this:
```
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2 -Wall")
```
It appears this was originally added with e7d11ba. These compiler flags seem to be gcc/clang specific, however the "-O2" is interpreted by MSVC as "/O2", which gets overridden by the specified "/Od" flag when configuring CMake as a debug build via:
```
cmake ..\libressl-3.3.4 -DCMAKE_BUILD_TYPE=Debug -G Ninja
```

One approach to solving this is changing the CMakeLists.txt line to:
```
if(NOT MSVC)
	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2 -Wall")
endif()
```

Additionally, I'm not sure the optimization flag should be changed like this, I think generally they should rely on the CMake system to set them. Generally, I believe CMake will set the appropriate optimization level provided the `CMAKE_BUILD_TYPE`.

busterb:
Fixes #683


botovq:
On OpenBSD this results in a significant performance regression,
presumably because there's no optimization:

```
$ time ./tests/bn_unit          # automake build
    0m06.25s real     0m05.87s user     0m00.49s system
$ time ./build/tests/bn_unit    # cmake with current defaults
    0m28.94s real     0m28.51s user     0m00.54s system
$ time ./O2/tests/bn_unit       # cmake with the -O2 added back
    0m06.16s real     0m05.77s user     0m00.41s system
```

busterb:
What about changing the default CMAKE_BUILD_TYPE from the current 'Debug', which sets -O0, to 'Release' which sets -O3, in a compiler-independent way.

Before:
```
$ time ./tests/bn_unit
./tests/bn_unit  6.26s user 0.08s system 95% cpu 6.656 total
```

After:
```
$ time ./tests/bn_unit
./tests/bn_unit  2.14s user 0.08s system 99% cpu 2.232 total
```

See #892

PR #892, opened 2023-07-27
set default build type to 'Release'

busterb:
A variation from https://www.kitware.com/cmake-and-the-default-build-type/ to default to enabling optimizations if no build type is specified.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.20.0)
+cmake_minimum_required (VERSION 3.16.4)
 if(MSVC)
 	cmake_policy(SET CMP0091 NEW)
 endif()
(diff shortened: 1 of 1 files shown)


botovq:
I haven't checked but I assume this change that adds `-DNDEBUG` and causes compiler warnings such as the below since `assert()` is neutered
```
/tmp/portable/tests/timingsafe.c:33:6: warning: unused variable 'cmp' [-Wunused-variable]
        int cmp = memcmp(bufone, buftwo, N);
```

busterb:
Correct:

//Flags used by the ASM compiler during DEBUG builds.
CMAKE_ASM_FLAGS_DEBUG:STRING=-g

//Flags used by the ASM compiler during MINSIZEREL builds.
CMAKE_ASM_FLAGS_MINSIZEREL:STRING=-Os -DNDEBUG

//Flags used by the ASM compiler during RELEASE builds.
CMAKE_ASM_FLAGS_RELEASE:STRING=-O3 -DNDEBUG

//Flags used by the ASM compiler during RELWITHDEBINFO builds.
CMAKE_ASM_FLAGS_RELWITHDEBINFO:STRING=-O2 -g -DNDEBUG

botovq:
We need to fix this before rolling tarballs.

Issue #893, opened 2023-08-06
3.8.0: CMake with MinGW fails to enable ASM support + ASM builds crashing


vszakats:
- Patch fixing the two build issues:

  1. MinGW detection to use `MINGW` variable instead of expecting specific `CMAKE_SYSTEM_NAME` values (this may be "Windows" even for MinGW builds.)
  2. Do not disable ASM support for MinGW.

  PR: #894

- The next issue is that ASM-enabled LibreSSL creates bad executables. They seems to crash before even starting.
It's a regression in 3.8.0. ASM-enabled 3.7.3 builds run fine.

vszakats:
The only difference between 3.7.3 and 3.8.0 ASM sources is the [addition](https://github.com/libressl/openbsd/commit/17bde8538e33793432c7952fdc0d5ddc780808e1) of `endbr64` instructions in 3.8.0.

OpenSSL [added support](https://github.com/openssl/openssl/commit/98ad3fe82bd3e7e7f929dd1fa4ef3915426002c0) for this in 2020, but it's not actually present in the x86_64 ASM source generated for the MinGW target (haven't tested others), as seen in a recent 3.0.8 build.

PR #894, opened 2023-08-06
cmake: fix to enable ASM support in MinGW builds

vszakats:
Use the built-in `MINGW` variable to detect MinGW, instead of relying on the `CMAKE_SYSTEM_NAME` variable being 
set to "MINGW" by the user. For Windows builds (including MinGW ones), this variable is expected to have the value "Windows". This fixes the logic to enable ASM support for MinGW by default.

Fix `crypto` lib build to let MinGW use the ASM support enabled above. Before this patch, it allowed ASM only for MSVC and disabled it for all other Windows builds.

After this patch ASM support is correctly enabled by default for MinGW builds when using CMake. Also bringing this in sync with autotools builds.

Ref #893


vszakats:
Rebased to master for greener CI results.

busterb:
Makes sense, thanks @vszakats

PR #895, opened 2023-08-06
Export CMake targets

FtZPetruska:
This PR adds support for exporting the CMake targets when building with CMake.

The advantages of using exported CMake configs over the FindLibreSSL modules are:
- Better knowledge of the library type. The target file stores information such as the build type (Release/Debug), and library type (static/shared). Moreover, when building a DLL on Windows, the DLL is properly set as the `IMPORTED_LOCATION`, and its implib as the `IMPORTED_IMPLIB`, this allows users to use CMake mechanisms such as `$<TARGET_RUNTIME_DLLS:tgt>` to automatically copy dependent DLLs on build/install.
- Better support for customisation. The `FindLibreSSL` module makes assumption about the headers and libraries location, but both locations can be overwritten by setting respectively `CMAKE_INSTALL_INCLUDEDIR` and `CMAKE_INSTALL_LIBDIR`. On the other hand, the exported targets keeps track of such customisation.

The goals I had were:
- Create a CMake package that was fully compatible with the FindLibreSSL module. This meant providing the same targets and result variables (`LibreSSL::*`, `LIBRESSL_*`).
- Make the package usable without installation. This means that a user can use `find_package(LibreSSL)`, from another project without installing LibreSSL to their system as such:
  ```
  cmake -S libressl -B libressl-build
  cmake --build libressl-build
  cmake -S my-project -B my-project-build -D LibreSSL_DIR=`pwd`/libressl-build
  ```

In order to test the changes, I added a simple CMake test project that calls `find_package(LibreSSL)`, checks the result variables, and builds and links an application for each library. This test is ran in CI for all three major platforms with the following three situations:
- Consuming the exported targets from the build directory
- Consuming the exported targets from an install using `CMAKE_PREFIX_PATH`
- Consuming the exported targets from an install using `LibreSSL_DIR`

Lastly, here are a few implementation notes:
- The additional CMake files are installed by default to `${CMAKE_INSTALL_LIBDIR}/cmake`, this can be customised by setting the cache variable `LIBRESSL_INSTALL_CMAKEDIR`.
- The public headers are copied into the build directory, this is especially useful when consuming the package without installing. It makes setting `LIBRESSL_INCLUDE_DIR` easier, and avoid leaking private headers.
- Github Actions wouldn't let me set the shell in the autogen step from the matrix. So I had to make two distinct steps to run autogen.sh depending on whether the runner is using Windows or not.

Thank you for your time reviewing this PR, let me know if you have any question!


busterb:
Can you please rebase this on the latest so it builds again?

PR #901, opened 2023-08-17
CI: add scheduled Fedora Rawhide builds

chipitsine:
Fedora Rawhide is shipped with the most recent gcc/clang, it is nice to test build on them from time to time


botovq:
We have a somewhat complicated method of merging PRs, so I missed that you marked this as a draft.
Let's see what happens. We can always revert.

chipitsine:
I'll send follow up PR, nevermind

PR #902, opened 2023-08-18
CI: fedora rawhide follow up

chipitsine:


.github/workflows/fedora-rawhide.yml
```diff
@@ -26,15 +26,11 @@ jobs:
       run: |
         ./autogen.sh
         ./configure
```
botovq, line 28: The `./configure` step is not needed for cmake builds.

(I also think `./autogen.sh` isn't really needed and `./update.sh` would be good enough, but let's leave it like that for now).

.github/workflows/fedora-rawhide.yml
```diff
@@ -26,15 +26,11 @@ jobs:
...
-      run: |
-         cd libressl-*
-         mkdir build-shared
-         cmake -GNinja -DBUILD_SHARED_LIBS=ON ..
+         CC=${{ matrix.cc }} cmake -GNinja -DBUILD_SHARED_LIBS=ON .
          ninja
+    - name: Test
+      run: |
          ninja test
+
```
botovq, line 36: zap trailinig empty line, please

botovq (changes requested):

botovq:
This workflow now fails to install the deps [with][1]

```
  file /usr/lib64/libimaevm.so.4.0.0 from install of ima-evm-utils-libs-1.6-0.fc41.x86_64 conflicts with file from package ima-evm-utils-1.5-4.fc40.x86_64
```

Before I waste a lot of time poking at an OS I don't know, do you perhaps know how to fix this?

Cc @joshuasing 

[1]: https://github.com/libressl/portable/actions/runs/9815355271/job/27104270522#step:4:355

chipitsine:
it happens from time to time on Rawhide, usually it is resolved by itself in several days
I'll check.

joshuasing:
> This workflow now fails to install the deps [with](https://github.com/libressl/portable/actions/runs/9815355271/job/27104270522#step:4:355)
> 
> ```
>   file /usr/lib64/libimaevm.so.4.0.0 from install of ima-evm-utils-libs-1.6-0.fc41.x86_64 conflicts with file from package ima-evm-utils-1.5-4.fc40.x86_64
> ```
> 
> Before I waste a lot of time poking at an OS I don't know, do you perhaps know how to fix this?
> 
> Cc @joshuasing

I don't use `dnf` very often, so I'm not sure what would cause this, but I can look into it.
If @chipitsine knows anything, that would be great, otherwise I'll also have a look to try and find a solution.

botovq:
On Sun, Jul 07, 2024 at 07:53:05AM -0700, Ilya Shipitsin wrote:
> it happens from time to time on Rawhide, usually it is resolved by itself in several days

Seems to have fixed itself (presumably some packaging error).

Thanks!

chipitsine:
the benefit of using Fedora |Rawhide is the most recent gcc and clang. The price however is stability

PR #909, opened 2023-09-04
Fix processor detection in CMakeLists.txt

jiegec:
`CMAKE_SYSTEM_NAME` is set to the OS name, e.g. Linux. It should check `CMAKE_SYSTEM_PROCESSOR` instead.

The regex is wrong: `[i?86|x86]` will matching anything containing `6`, so it will take the wrong branch for many ISAs, e.g. riscv64 and even ARM64 when it is uppercased. It eventually leads to wrong DES encryption due to wrong `sizeof(DES_LONG)`.

Handle cases when `CMAKE_SYSTEM_PROCESSOR` is upper-cased, e.g. in windows.


jiegec:
CI failed in windows:

```
Building Custom Rule D:/a/portable/portable/crypto/CMakeLists.txt
  Assembling D:\a\portable\portable\crypto\aes\aes-masm-x86_64.S...
D:\a\portable\portable\crypto\aes\aes-masm-x86_64.S(1840): error A2189: invalid combination with segment alignment : 64 [D:\a\portable\portable\build\crypto\crypto_obj.vcxproj]
D:\a\portable\portable\crypto\aes\aes-masm-x86_64.S(2230): error A[21](https://github.com/libressl/portable/actions/runs/6068067965/job/16460548254?pr=909#step:6:22)89: invalid combination with segment alignment : 64 [D:\a\portable\portable\build\crypto\crypto_obj.vcxproj]
D:\a\portable\portable\crypto\aes\aes-masm-x86_64.S(2624): error A2189: invalid combination with segment alignment : 64 [D:\a\portable\portable\build\crypto\crypto_obj.vcxproj]
```

I don't have Windows environment to test.

vszakats:
Another case is Universal macOS builds, setup with e.g. `CMAKE_OSX_ARCHITECTURES=arm64;x86_64`.

I don't think the current build layout can handle that, and matching one
of the platforms seems wrong as well. Possibly ASM should be disabled
with multiple values there. I haven't made tests though.

It'd be useful to make that work out of the box. For that, the CPU 
selection should probably be delegated to the C preprocessor by 
#including the necessary ASM code into C at built-time, and keep the
set of source files the same, regardless of target CPU.

busterb:
Hi, thanks for the PR. I think what you are doing here is correct; the Windows failure is due to it actually fixing the detection of processor and enabling MASM assembly in the build, which is coincidentally broken. I'll add a patch to this to work around that and fix the MASM in another patch set so this applies cleanly on its own.

PR #912, opened 2023-09-12
MSVC: Fixes from vcpkg

FtZPetruska:
MSVC currently has two patches for its LibreSSL port.

[The first patch](https://github.com/microsoft/vcpkg/blob/master/ports/libressl/0001-enable-ocspcheck-on-msvc.patch) enables building `ocspcheck` with MSVC. This requires:
- `getopt`, which is already included in `libcrypto`. Adding `optarg` and `optind` to the DLL exports is necessary.
- `ftruncate`, which is already included in `libtls`. Adding it to the DLL exports is necessary.
- Defining `STDIN_FILENO` in the `unistd.h` compat header.

[The second patch](https://github.com/microsoft/vcpkg/blob/master/ports/libressl/0002-suppress-msvc-warnings.patch) handles the [Spectre vulnerability mitigation](https://learn.microsoft.com/en-us/cpp/build/reference/qspectre?view=msvc-170). The warning [C5045](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5045?view=msvc-170) triggers whether `/Qspectre` is used or not and warns about code that may be affected by this vulnerability.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,6 +163,8 @@ if(MSVC)
 			"C4706" # assignment within conditional expression
 			"C4996" # The POSIX name for this item is deprecated.
 				# Instead, use the ISO C and C++ conformant name
+			"C5045" # Compiler will insert Spectre mitigation for memory
+				# load if /Qspectre switch specified
 		)
 	elseif(CMAKE_C_COMPILER_ID MATCHES "Intel")
 		add_definitions(-D_CRT_SUPPRESS_RESTRICT)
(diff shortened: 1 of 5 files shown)


busterb:
Hi, thanks for the PR. I think for now would like to leave the Spectre mitigation off by default, not because we don't want it, but it seems that the VS environment being built in needs to have additional libraries installed, and the check for version isn't sufficient to check for the libraries as well.

busterb:
I cherry-picked the oscpcheck fixes in c4bb6b79ea36c05ccfe3033174a7950ac6a51609, thanks!

 Let's discuss the spectre mitigations in another PR. I'm not sure if we really want this to happen automatically or intentionally on behalf of the builder.

PR #913, opened 2023-09-12
Cmake improvements

marc-groundctl:
`check_required_components` does not take a list of components; it takes the package name and determines if a component is available by checking the `<package>_<component>_FOUND` variable.  Without this change, `find_package(LibreSSL REQUIRED COMPONENTS Invalid)` would succeed despite that component not existing.

This change also makes `LIBRESSL_FOUND` mirror the `LibreSSL_FOUND` variable that cmake uses internally to check if the package exists, and changes the `install` source paths of the cmake files to match the ones they were configured with.


busterb:
Thanks for the PR @marc-groundctl

PR #930, opened 2023-11-04
cmake: fix default `TLS_DEFAULT_CA_FILE` value in `libtls` for Windows

vszakats:
Matching the logic used for `libcrypto` (and in autotools builds
for both `libcrypto` and `libtls`), set the default
`TLS_DEFAULT_CA_FILE` value to `C:/Windows/libressl/ssl/...`,
replacing the current, potentialy world-writable filename.


busterb:
This is in a few different places. Wondering if we could better condense this logic to a common place.

vszakats:
Good point! I've made a commit that also fixes this for `nc` and `ocspcheck` while moving up the defaulting logic to the root `CMakeLists.txt`.

busterb:
This looks better, thanks!

PR #931, opened 2023-11-04
slim down `libtls`

vszakats:
By dropping copies of `libcrypto` and `libssl` lib objects from it.
This brings back its size to the normal ~100KB from ~4MB.

autotools already worked like this.


busterb:
Thanks for the PR @vszakats

PR #932, opened 2023-11-04
slim `tls-static` library (used in tests)

vszakats:
Instead of including a full copy of libcrypto and libssl in libtls-static, link existing libcrytpo-static and libssl-static to the test targets.

This wasn't causing any issue, just unnecessarily duplicating a lot of objects.


busterb:
Thanks @vszakats

PR #936, opened 2023-11-07
cmake: fix arm64 to not trigger armv4 ASM

vszakats:
Reported here:
https://github.com/libressl/portable/pull/935#issuecomment-1798345787


vszakats:
test: https://github.com/curl/curl-for-win/actions/runs/6784132320

busterb:
Thanks @vszakats

PR #941, opened 2023-11-08
ci: shrink and improve Linux workflows

joshuasing:
Add a new `linux` workflow that replaces cross_test, linux_ci, linux_ci_asan, and linux_ci_asan_noasm.

Changes:
 - Create new `linux.yml` workflow
 - Restrict GitHub Action token permissions to `contents: read`
 - Pin `actions/checkout` to `v4`
 - Use more common naming schemes

The new workflow will run:
 - When a commit is pushed to a branch.
 - When a pull request is created/updated.
 - On a fixed schedule of `0 0 * * *` (daily at 00:00).


.github/workflows/linux.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: "Run tests"
+        run: ./scripts/test
+        env:
+          ARCH: "${{ matrix.arch }}"
+          CC: "${{ matrix.compiler }}"
+
+  # Test ASAN with and without ASM enabled.
+  test-asan:
+    name: "ASAN (ASM${{ matrix.asm == 'OFF' && ' OFF' || '' }})"
```
4a6f656c, line 37: Using "no-asm" and "asm" might be clearer...

4a6f656c (approved):
LGTM

busterb:
Going from 33 -> 25 CI checks, did we lose anything important here, or was it redundant?

4a6f656c:
> Going from 33 -> 25 CI checks, did we lose anything important here, or was it redundant?

Good catch - looks like cross_ci and linux_test were running with both ubuntu-20.04 and ubuntu-22.04, while this is now only running with ubuntu-latest... we'll want to restore that.

joshuasing:
Oops, I have switched back to running on both `ubuntu-20.04` and `ubuntu-22.04`.
Additionally, I have made the suggested change to the ASAN name.

busterb:
Looks good, thanks for the updates.

PR #946, opened 2023-11-10
ci: add --verbose flag to rust-openssl cargo test

joshuasing:
Add `--verbose` flag to the `cargo test` command in the `rust-openssl` workflow to allow for more visibility.

Requested in https://github.com/libressl/portable/pull/944#discussion_r1389316062


botovq:
As I said, this isn't running the tests against LibreSSL. There is [` --cfg ossl300` ][1]. I'm quite sure  you need to export those shell variables.

[1]:     https://github.com/libressl/portable/actions/runs/6831414608/job/18580856558#step:5:634

PR #947, opened 2023-11-11
Fix rust-openssl regress

botovq:
Change around the order slightly. It seems more logical to apply the patch first, then export the env.


botovq:
This properly [links against libressl][1]


[1]:     https://github.com/botovq/libressl-portable/actions/runs/6835799514/job/18589995738#step:5:122

PR #949, opened 2023-11-18
remove additional endbr64 patch, add tests for macos-11

botovq:


busterb:
The only issue I had with this earlier is that the macos-11 builder still doesn't show any issues with building with endbr64 out of the box. 

Supposedly you can select alternate compilers, but I didn't look too much further, and it really depends on the assembler anyway: https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md

PR #951, opened 2023-11-28
disable periodic builds of broken CI workflows

busterb:
Neither of these has built in over a month.

Solaris looks to be an upstream timeout issue (seems the hack it uses may not work anymore).
https://github.com/libressl/portable/actions/runs/6999379749

Not sure if cifuzz is really needed either, since we're still being tested through clusterfuzz actively.
https://github.com/libressl/portable/actions/runs/7012733325

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,10 +1,5 @@
 name: CIFuzz
 
-on:
-  workflow_dispatch:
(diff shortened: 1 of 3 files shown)


botovq:
Thanks. Could you also remove the solaris_ci badge from README.md, please?

joshuasing (approved):

PR #953, opened 2023-12-05
Fix build on mips

maaaat:


pcercuei:
I was facing the same issue and came up to the same fix, so ACK.

botovq:
We will roll a new stable release relatively soon and will include this
fix.

Thanks!

PR #959, opened 2023-12-10
delete ssl_algs.c references

vszakats:


botovq:
I don't understand. https://github.com/libressl/portable/commit/a332334ec11c68ca15613158b94fc72528564c38

vszakats:
Sorry for the noise, just a case of outdates sources.

botovq:
No problem. I'm more confused about why GitHub doesn't notice. But then again we are using merge commits in this repo, everything just has to be utterly confusing.

Issue #928, opened 2023-11-03
Can we prefix compat symbols to avoid symbol clashes in static links?


botovq:
While we no longer export the compat symbols from the dynamic libraries, there is still an issue with static linking, see e.g., https://github.com/curl/curl/issues/12257

Can we leverage a trick like the one used in the hidden directories to prefix these symbols transparently?

botovq:
I wonder if we can do something like this in the various compat headers:

 ```C
#ifndef HAVE_ARC4RANDOM
#define arc4random _libressl_arc4random
uint32_t arc4random(void);
#endif
```

This way we don't need to litter the main source code with prefixes and the symbols should no longer leak out of the libraries, even with static linking.

botovq:
On Fri, Nov 10, 2023 at 06:11:19AM -0800, Bob Beck wrote:
> Is it really only the arc4random that becomes a problem here now that some systems have it? 

All compat symbols are a problem. And there may be internal symbols that
could also potentially clash.

vszakats:
Another non-prefixed symbol, `strtonum` is causing a number of issues on macOS:

1. autotools detects it even if the OS target version doesn't offer it. When referenced
  from the code, it results in https://github.com/libressl/portable/issues/910#issuecomment-1708829195.
  ```
  note: 'strtonum' has been marked as being introduced in macOS 11.0 here, but the deployment target is macOS 10.9.0`
  ```
  The workaround for that is to suppress detection with `export ac_cv_func_strtonum='no'`.
  That shifts the problem into the next ones.

2. cmake doesn't make an attempt to auto-detect `strtonum`, by default assuming
  that the function isn't provided. But since it is in fact provided via `stdlib.h` (with
  the caveat above), these warnings appear when using it:
  ```
  libressl/crypto/x509/x509_addr.c:1645:17: warning: 'strtonum' is only available on macOS 11.0 or newer [-Wunguarded-availability-new]
                          prefix_len = strtonum(s + i2, 0, 8 * length, &errstr);
                                       ^~~~~~~~
  libressl/crypto/../include/compat/stdlib.h:43:11: note: 'strtonum' has been marked as being introduced in macOS 11.0 here, but the deployment target is macOS 10.9.0
  long long strtonum(const char *nptr, long long minval,
            ^
  libressl/crypto/x509/x509_addr.c:1645:17: note: enclose 'strtonum' in a __builtin_available check to silence this warning
[...]

PR #961, opened 2023-12-11
Prefix some compat symbols with libressl_

botovq:
See #928. This isn't a full fix, but should remove much of the friction already.


botovq:
I was hoping that we could backport this to stable, but unfortunately the `EXTRA_EXPORT` things for CMake means that this amounts to a libcrypto major bump on any platform that needs extra symbols.

joshuasing (approved):

busterb:
I'm merging fixes to stable, noting that this PR will be skipped.

PR #972, opened 2023-12-14
cmake: stop exporting compat functions

### LINKED ISSUE #969: Why do CMake builds still have EXTRA_EXPORTs?
I would love for there to be a change matching

https://github.com/libressl/portable/commit/54b31ce5ec82248bba32f1a7c49e66e088cb1807

for CMake builds.

See also https://github.com/libressl/portable/commit/f7bed14f22b70198b4ec56af6d489f69b3df805a#r134917738 for context.

vszakats:
Attempt to avoid `EXTRA_EXPORT` and `LIBTLS_EXTRA_EXPORT`, by doing something
similar to what autotools builds do already.

Fixes #969

---

~~Locally I could only test a 3.8.2-based iteration and one that didn't tackle `LIBTLS_EXTRA_EXPORT`. CI should tell more.~~


crypto/CMakeLists.txt
```diff
@@ -798,105 +781,88 @@ if(NOT HAVE_GETPROGNAME)
...
 endif()
 
 if(NOT HAVE_SYSLOG_R)
 	set(CRYPTO_SRC ${CRYPTO_SRC} compat/syslog_r.c)
 endif()
 
 if(NOT HAVE_TIMEGM)
-	set(CRYPTO_SRC ${CRYPTO_SRC} compat/timegm.c)
-	set(EXTRA_EXPORT ${EXTRA_EXPORT} timegm)
+	set(COMPAT_SRC ${COMPAT_SRC} compat/timegm.c)
```
botovq, line 823:
```suggestion
        # XXX - Remove after next bump once libtls, ocspcheck, and asn1time are fixed.
	set(COMPAT_SRC ${COMPAT_SRC} compat/timegm.c)
```

botovq (approved):
Amazing. Many thanks!

I have come up with a subset of this diff, but haven't gotten quite as far. All the changes make sense. I wonder if we can use a relatively recent API that will likely be needed everywhere such as recallocarray in place of empty.c.

busterb:
Note: skipping this PR for stable

PR #980, opened 2023-12-19
ci: add concurrency groups to workflows

joshuasing:
This pull request adds concurrency groups to all existing GitHub Actions workflows.

When a pull request is created, or a commit is pushed to a branch, all workflows related to that event are triggered. However, when subsequent commits are pushed to that pull request/branch, the workflow runs may be delayed until the previous workflow runs complete, due to GitHub's runner concurrency limit.

By adding concurrency groups to the workflows, we can automatically cancel all pending or running workflow runs related to previous commits, allowing workflow runs on newer commits to start faster.

This works by grouping each workflow run by the workflow name and git reference (branch/pull request). When concurrent workflow runs with the same group are queued, this will cause any previous pending or running jobs to be cancelled automatically (`cancel-in-progress: true`).

Related: https://docs.github.com/en/actions/using-jobs/using-concurrency

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -3,6 +3,10 @@ name: "Android"
 
 on: [push, pull_request]
 
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+
 jobs:
   test:
     name: "Test ${{ matrix.name }}"
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -3,6 +3,10 @@ name: CIFuzz
 on:
   workflow_dispatch:
 
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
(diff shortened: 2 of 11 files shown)


botovq:
I see how this can be an improvement while stacking up commits on a PR.

Is the flipside of this that if I break, say, some MinGW workflow, all
other runs are cancelled or does breakage not count as cancellation?

If it does count, I'm not sure this will be a quality of life improvement.
If some base commit breaks some CI, I need to have a complete picture of
the extent of the breakage as quickly as possible.

joshuasing:
> I see how this can be an improvement while stacking up commits on a PR. Is the flipside of this that if I break, say, some MinGW workflow, all other runs are cancelled or does breakage not count as cancellation? If it does count, I'm not sure this will be a quality of life improvement. If some base commit breaks some CI, I need to have a complete picture of the extent of the breakage as quickly as possible.

If a workflow run fails, all other workflows should continue.
The concurrency groups only cause existing workflow runs to be cancelled when the workflows are re-triggered with the same group (e.g. if a new commit is pushed to the branch or pull request)

botovq:
Just wanted to say thanks for this change. It's pretty great for people like me who tend to push before they think hard.

PR #984, opened 2023-12-26
Avoid empty translation unit

botovq:


botovq:
Ok, this hack [didn't work][1]. A comment isn't enough to count. We could probably use a typedef, but, really, there must be a way to convince cmake to do this without such gross workarounds.

[1]: https://github.com/libressl/portable/actions/runs/7333397364/job/19968780861?pr=984#step:6:29

busterb:
Maybe there's a better fix overall for #776 ?

botovq:
Yes, it's related, but to be honest, I was just playing around a bit since I was bored...

I think we might be able to build a list of compat source files and only create the compat libraries if that list is non-empty.

PR #988, opened 2023-12-31
cmake: disable default NDEBUG differently

vszakats:
This method allows to set it via custom flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,8 +62,13 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
 	  STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
 endif()
 
-# Enable asserts regardless of build type
-add_definitions(-UNDEBUG)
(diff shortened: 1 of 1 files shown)


vszakats:
`-UNDEBUG` seems to hide valid build issues at compile-time, so an possibly
better solution would be to just delete the logic touching this macro and leaving
it to the CMake default.

busterb:
Thanks. I was just looking at build logs for another diff, being slightly annoyed there wasn't a better way to override NDEBUG without a bunch of warnings.

vszakats:
Is there something to do, or any pending concern with this PR to get merged?

vszakats:
This patch was merged, yet surprisingly, it's missing from master and both from 3.8.3 and 3.9.0.

Was it intended?

busterb:
I wonder if this got merged to the github mirror only, then was overwritten when we synced up. I only see it referred to in the PR branch in the actual source tree, and there wasn't a revert that I can see.

vszakats:
Yes, that's possible.

I've opened #1016 to try one more time. Or the previous patch 
can also be re-applied, whichever is more convenient.

PR #989, opened 2024-01-02
cmake: disable ASM for Windows ARM64

### LINKED ISSUE #987: Windows ARM64 assert in `bn_div.c`
Getting this fatal error when trying to do any HTTPS transfer with LibreSSL + curl on Windows ARM64:
```
Assertion failed: (i == BN_BITS2) || (h <= (BN_ULONG)1 << i), file \crypto\bn\bn_div.c, line 96
```

Using this binary:
https://curl.se/windows/dl-8.5.0_3/curl-8.5.0_3-win64a-mingw.zip

Another issue is that the LibreSSL `CMakeLists.txt` force-disables `NDEBUG`.
asserts aren't expected in release builds. CMake enables `NDEBUG` in release
mode to avoid them by default.

vszakats:
With ASM support the builds either exit with an assert or hang (with asserts disabled).

autotools doesn't enable ASM for Windows ARM64, so it's not affected.

Fixes #987


joshuasing (approved):

botovq:
There is no assembly on aarch64 to speak of so I don't think doing this
changes much. However, it feels like there is a different issue. Maybe
the windows special with having 32-bit longs is confusing some #ifdefs?

@busterb, could you take a look?

vszakats:
There is a long list of `-Wasm-operand-widths` compiler warnings (which I did miss throughout all the testing!)
and they do look like coming from inline ASM code in `crypto/bn/arch/aarch64/bn_arch.h`:
https://github.com/curl/curl-for-win/actions/runs/7224401784/job/19685602186#step:3:9017

Wrongly detected integer sizes in `openssl/bn.h` might be a reason.

vszakats:
FWIW, after:
https://curl.se/windows/dl-8.5.0_4/curl-8.5.0_4-win64a-mingw.zip (working)
and before binaries:
https://curl.se/windows/dl-8.5.0_3/curl-8.5.0_3-win64a-mingw.zip (failing with assert)

(though this disables ASM differently; without patching LibreSSL.)

busterb:
Hey folks, you're right about the LLP64 nature of Windows 64-bit being the core issue here. I remember it being difficult to untangle last time I looked, but this seems like a reasonable workaround until it can be addressed more generally.

PR #990, opened 2024-01-09
Replace uses of endbr64 with _CET_ENDBR from cet.h (test patch)

busterb:
cet.h is needed for other platforms to emit the relevant .gnu.properties sections that are necessary for them to enable IBT. It also avoids issues with older toolchains on macOS that explode on encountering endbr64.

This also fixes an data alignment issue that occur when building on Windows where .section directives get a fixed 4 or 8 byte alignment, but then the section alignment is overridden on the next line, causing an assembler error.


busterb:
Failed checks here are due to github having a moment https://www.githubstatus.com/

patches/asm/masm-align-64.patch
busterb, line 1: @botovq how does this look? I know, it's almost readable this way, but I couldn't figure out how to add more ternary operators.

botovq: If it works, I'm fine with this. Thanks.

botovq: Very cool. Thanks for pushing this over the line!

botovq (approved):

botovq:
I assumed you left the endbr64 patch in here for ease of cherry-picking
to stable. It should however be committed to -current in CVS. Will you
take care of this or should I?

PR #995, opened 2024-01-29

chipitsine:
couple of jobs were missed previously

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -46,6 +46,7 @@ jobs:
   test-asan:
     name: "ASAN (${{ matrix.asm == 'ON' && 'asm' || 'no-asm' }})"
     runs-on: "ubuntu-latest"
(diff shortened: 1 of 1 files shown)


botovq:
On Mon, Jan 29, 2024 at 01:03:35PM -0800, Ilya Shipitsin wrote:
> couple of jobs were missed previously

Thanks. I guess I should delete a couple more files to make sure we
caught them all now. Will do so soon, I think. :)

chipitsine:
what files do you want to delete ? (just curious)

botovq:
There are stil a few files with only [2 or 3 lines of code][1].

[1]: https://github.com/libressl/openbsd/blob/30d39d87f7d79b965a5739bbd0f60511c67b140f/src/lib/libcrypto/pkcs12/p12_p8d.c#L66

chipitsine:
honestly, I thought of "empty.c"

ok, I'll try to delete it myself

PR #996, opened 2024-01-31
crypto: do not create empty.c anymore

chipitsine:


joshuasing (approved):

botovq:
I am not sure about this. This relies on the fact that all supported
platforms will use some compat code for libcrypto, otherwise cmake
will be unhappy due to missing object files for libcompat.

While this appears to be true based on the CI results and testing this
on OpenBSD, I believe at least for OpenBSD this is a bug in the compat
code: it detects missing getauxval support - which is correct but not
needed - and missing b64_pton - which is incorrect and probably due to
naming the symbol __b64_pton, and you need resolv.h to find b64_pton.

Rather than removing it outright, I believe we need to handle this
correctly in cmake. Presumably we need an empty list of compat files
and handle libcompat only if that list isn't empty.

busterb:
This was originally a workaround to creating static libraries with XCode. See here: https://github.com/libressl/portable/pull/776

Probably worth revisiting the issue.

chipitsine:
Thanks, I'll check it out

chipitsine:
the idea of removing empty.c is to get rid of those warnings

https://github.com/libressl/portable/actions/runs/7869794632/job/21469627057#step:6:29

and once we'll be able to switch windows builds to "all warning are errors"

but there's no rush

busterb:
Sounds like what we need to do here next is verify if this is still needed for Xcode static builds. If it is, either find another solution, or guard it for that specific case.

chipitsine:
I plan to try something next weekend

chipitsine:
I reviewed https://github.com/libressl/portable/pull/776
it introduces "dummy.c", changes were reverted lately. currently that file is not used.

I think we can manipulate with empty.c removal without taking that into account.

also, I did not find any repro step how to build "iOS static build with XCode generator", it might be good to add such a build to CI/CD

chipitsine:
well. there's no rush.
I'll try to find out what is "xcode ios cmake...."

busterb:
Verified that Xcode still has trouble generating static libraries with the default CMake build if empty.c is removed. Here's what results:

image

chipitsine:
I do not own Mac, I can try something on CI using xcodebuild.
which steps did you take before building ?

пн, 18 мар. 2024 г. в 03:54, Brent Cook ***@***.***>:

chipitsine:
as far as I understand, it makes sense to add such "iphone static xcode" builds to CI first to be able to catch regressions

busterb:
Absolutely. I can submit a PR with this so we can catch it automatically.

chipitsine:
nevermind, I've found https://github.com/build-xcframeworks/libressl

пн, 18 мар. 2024 г. в 21:49, Brent Cook ***@***.***>:

Issue #266, opened 2017-01-26
tls_close causes exception on Windows


vsajip:
Given the following sequence of events:

1. A connected socket is associated with a context using `tls_connect_socket`.
2. The socket is closed (after handshake).
3. `tls_close` is called on the context.

On Windows (but not Linux) the program crashes in the `tls_close` call. The stack trace indicates that a write is being attempted to the closed socket, which is causing Windows to  abort the program with `An invalid parameter was passed to a function that considers invalid parameters fatal`.

Stack trace:
 	ucrtbased.dll!00007ff92c379fb8()
 	libcrypto.dll!posix_write(int fd, const void * buf, unsigned __int64 count) Line 188
 	libcrypto.dll!sock_write(bio_st * b, const char * in, int inl) Line 154
 	libcrypto.dll!BIO_write(bio_st * b, const void * in, int inl) Line 252
 	libssl.dll!ssl3_write_pending(ssl_st * s, int type, const unsigned char * buf, unsigned int len) Line 813
 	libssl.dll!do_ssl3_write(ssl_st * s, int type, const unsigned char * buf, unsigned int len, int create_empty_fragment) Line 789
 	libssl.dll!ssl3_dispatch_alert(ssl_st * s) Line 1414
 	libssl.dll!ssl3_send_alert(ssl_st * s, int level, int desc) Line 1400
 	libssl.dll!ssl3_shutdown(ssl_st * s) Line 2544
 	libssl.dll!SSL_shutdown(ssl_st * s) Line 1029
 	libtls.dll!tls_close(tls * ctx) Line 654

The `fd` in the `posix_write` call is the same as the fd of the connected socket passed to `tls_connect_socket`. This doesn't seem like desirable behaviour for `tls_close`, or is there an expectation that the socket must be closed later?

busterb:
Do you have an example program that you can share?

kinichiro:
It appears to this commit https://github.com/libressl-portable/openbsd/commit/6be5c4c0476b00de370b0455be9ed66dc8a6e68f relates to this issue.
Can you confirm this solve your issue ?

vsajip:
> Do you have an example program that you can share?

See [this Gist](https://gist.github.com/vsajip/7e4ba0f06822b1b46815de5e48826c2c).

When run, this program prints the issuer information and then crashes:

![Failure dialog](http:/i.imgur.com/TcphXXe.png)

> Can you confirm this solve your issue ?

[...]

PR #997, opened 2024-02-04
Windows: Prevent assertion pop-up up when using Debug compiled libressl.

RandomInEqualities:
Hi, 

I tried to run the test suite in Debug mode on windows and had an assertion window pop. 

It happens when running the signertest, or the test project in  https://github.com/libressl/portable/issues/266. This was fixed in afcd4be8a72a for a release compiled library. To prevent the issue in debug mode, it looks like it is necessary to  also disable the assertion window popup. 

With this all tests pass when compiling and running them with a Debug, Release or RelWithDebInfo CMake build on windows (for me).


RandomInEqualities:
It is this one that pops up:

image

when for example running the signertest.

PR #998, opened 2024-02-04
Windows: Improve check for endianness when using Visual Studio.

RandomInEqualities:
Hi,

I noticed that ´__BYTE_ORDER__´ was not defined when compiling with visual studio, causing it to detect everything as little endian. This adds a cmake build check for endianness, if you think it makes sense to do this.

Visual studio does not define ´__BYTE_ORDER__´  so all architectures were detected as LITTLE_ENDIAN since both ´__BYTE_ORDER__´ and ´__ORDER_LITTLE_ENDIAN__´ would evaluate to 0 and compare equal. This updates the checks to use CMakes detection of endianness, with a hard error, if this also fails.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,6 +142,15 @@ endif()
 if(MSVC)
 	add_definitions(-Dinline=__inline)
 	message(STATUS "Using [${CMAKE_C_COMPILER_ID}] compiler")
+
+	include(TestBigEndian)
+	TEST_BIG_ENDIAN(HAVE_BIG_ENDIAN)
+	if(HAVE_BIG_ENDIAN)
+		add_definitions(-DHAVE_BIG_ENDIAN)
+	else()
+		add_definitions(-DHAVE_LITTLE_ENDIAN)
+	endif()
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -186,6 +186,14 @@ if(MSVC)
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -MP -W4 ${MSVC_DISABLED_WARNINGS_STR}")
 endif()
 
+include(TestBigEndian)
+TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
+if(IS_BIG_ENDIAN)
+	add_definitions(-DHAVE_BIG_ENDIAN)
+else()
+	add_definitions(-DHAVE_LITTLE_ENDIAN)
+endif()
```
vszakats, line 195: Suggesting to guard this with `if(MSVC)`, to avoid doing this check unnecessarily with gcc/clang toolchains.

RandomInEqualities: Right yes it makes sense. Thank you. It sounds better to separate it, so this is the fallback path. I have update the changes with it, if it looks better!

PR #1000, opened 2024-02-13
test windows builds with shared libs enabled

busterb:
Noticed a failure in appveyor with shared libs enabled, do the same in workflows

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -13,7 +13,7 @@ concurrency:
 
 jobs:
   test:
-    name: "${{ matrix.os }}/${{ matrix.arch }} (${{ matrix.generator }})"
+    name: "${{ matrix.os }}/${{ matrix.arch }}/shared ${{ matrix.shared }} (${{ matrix.generator }})"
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -13,7 +13,7 @@ concurrency:
 
 jobs:
   test:
-    name: "${{ matrix.os }}/${{ matrix.arch }} (${{ matrix.generator }})"
+    name: "${{ matrix.os }}/${{ matrix.arch }}/shared ${{ matrix.shared }} (${{ matrix.generator }})"
```
joshuasing, line 16: I think formatting this as  `windows-2022/ARM64 (Visual Studio 17 2022, shared)`/`windows-2019/Win32 (Visual Studio 16 2019` could make it easier to understand. Maybe something like this would work?
```suggestion
    name: "${{ matrix.os }}/${{ matrix.arch }} (${{ matrix.generator }}${{ matrix.shared == 'ON' && ', shared' || '' }})"
```

busterb: Sure, let's give it a try.

PR #1005, opened 2024-02-21

chipitsine:
second attempt to land https://github.com/libressl/portable/pull/995  :)


botovq:
Thanks. Looks like this is the last one...

PR #1007, opened 2024-02-25
fixup masm asm mneumonics in perl generator instead of update.sh

busterb:
removes dependency on cpp. Testing with newer masm, not running into the #include issues that were found years ago. Maybe we're good to remove this now.


busterb:
argh, alright this needs more thought. dealing with the lack of preprocessor in masm is a pain I'm only just re-remembering.

PR #1010, opened 2024-03-03
Stable cherry picks for OPENBSD_7_4 branch

busterb:
Cherry picked fixes from master to go into the next stable release.


botovq:
The libressl_ changes rename exported symbols, I don't think we should
be doing this in a -stable release. For example, a project may have
picked up our compat arc4random in a configure check and this might stop
working with the updated lib.

Do we really need to backport all this? The next release will be
relatively soon and, as far as I remember, there were only a few
complaints because of endbr64 breaking on old OS.

botovq:
Since you've already done the work, I'm not against it, but please drop
aa7584a and d36aef9 (and compare the nm -g output with and without the
cherry-pick to make sure we don't change the exported symbols.)

busterb:
Removed aa7584a and d36aef9 and all the dependent diffs as well.

Since many of the fixes are improvements to the test infra, I'd rather have the tests work the same on both branches. Also, the we needed some other changes to even build on OpenBSD properly with the endbr64 updates, it was easier to be more the same than different between these branches.

botovq:
Yeah, totally makes sense. I'm fine with this though I haven't had time
to check the details.

patches/opensslv.h
```diff
@@ -0,0 +1,18 @@
+/* $OpenBSD: opensslv.h,v 1.77 2023/09/20 11:42:25 tb Exp $ */
+#ifndef HEADER_OPENSSLV_H
+#define HEADER_OPENSSLV_H
+
+/* These will change with each release of LibreSSL-portable */
+#define LIBRESSL_VERSION_NUMBER 0x3080300fL
+/*                                    ^ Patch starts here   */
+#define LIBRESSL_VERSION_TEXT   "LibreSSL 3.8.3"
```
botovq, line 8: Should this not be a diff rather than the header itself?

botovq (approved):

joshuasing (approved):

patches/opensslv.h
botovq: Ah, perfect. Thanks!

PR #1012, opened 2024-03-07
put compat getpagesize into the right object list

busterb:
Was getting this linker error building shared libraries on Windows:

```
Creating library C:/projects/portable/build/ssl/Release/ssl.lib and object C:/projects/portable/build/ssl/Release/ssl.exp
recallocarray.obj : error LNK2019: unresolved external symbol getpagesize referenced in function
```

getpagesize should be in compat rather than crypto object file list.

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -734,7 +734,7 @@ if(NOT HAVE_GETOPT)
 endif()
 
 if(NOT HAVE_GETPAGESIZE)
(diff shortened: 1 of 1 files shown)


busterb:
I think so, if they are used outside of libcrypto. I only noticed this in appveyor build logs, haven't reproed locally yet.

botovq:
getprogname() is certainly used outside of libcrypto (e.g., tls_verify.c). I think syslog_r is needed for OPENSSL_showfatal() and getentropy() shims (which are in libcompat), so I think they both should go into COMPAT_SRC.

PR #1019, opened 2024-03-17
CI: setup kernel entropy to work with asan

chipitsine:


botovq:
Thanks a lot!

busterb:
Thanks as well, I was scratching my head over this this morning.

botovq:
Seems this is no longer needed.

PR #1020, opened 2024-03-18
Add `tests/bio_dump*` and `tests/x509_algor*` to gitignore

nak3:
As 37d868b and 04fa997 added two new tests, this patch adds them to the gitignore to prevent from pushing copied files.


botovq:
Thank you

PR #1022, opened 2024-03-18
Add tests for macos-14

nak3:
This patch adds test for macos-14.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -25,7 +25,7 @@ jobs:
         arch: ["arm64", "x86_64"]
     steps:
     - name: "Install required packages"
-      run: brew install automake
+      run: brew install automake libtool
 
     - name: "Checkout repository"
       uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


nak3:
Note - macos-14 needs to install libtool otherwise it got an error as https://github.com/nak3/portable/actions/runs/8323041680/job/22771906307

joshuasing (approved):
Thanks!

Related information that could become useful later:
https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/
https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md

botovq:
I'm not super keen on making the macos CI  even slower than it already is. Can we perhaps split this job into two that run in parallel or drop the oldest?

nak3:
Alright, I dropped the oldest (macos-11).

botovq:
Thanks. @joshuasing any opinion on this?

joshuasing:
> Thanks. @joshuasing any opinion on this?

@botovq I think this is a good change, and I have approved the pull request.

botovq:
@joshuasing You approved before macos11 was dropped and I can't see a renewal of the approval in the timeline, hence my question.

PR #1026, opened 2024-03-20
update.sh: use shallow clone to spare time/bandwidth

vszakats:


botovq:
I'm happy with this. Any particular reason for the choice of 8 commits?

vszakats:
No strong reason. I picked this back then when experiencing issues with 1 in CI jobs
(can't recall the specifics). It also helps sometimes to peek into the most recent
commit history on a local machine. Perf and size-wise there is no difference
between e.g. 8 and 1.

joshuasing (approved):

busterb:
Maybe a small tweak, replace the whole block with this:

```
LIBRESSL_GIT_OPTIONS="${LIBRESSL_GIT_OPTIONS:- --depth=8}"
LIBRESSL_GIT="${LIBRESSL_GIT:- https://github.com/libressl}"
git clone $LIBRESSL_GIT_OPTIONS $LIBRESSL_GIT/openbsd
```

Then if someone is relying on something else, or the previous behavior, they can just define LIBRESSL_GIT_OPTIONS=" " in their environment.

vszakats:
Updated with the lines above.

joshuasing (approved):

busterb:
This seems to be breaking the ability to make a branch release; investigating now

PR #1031, opened 2024-03-25
Add libssl and libcrypto to libtls's Libs.private

alyssais:
Since d193f43 ("slim down `libtls`"), statically linking libtls requires also linking libssl and libcrypto, so express that in the pkg-config file.


joshuasing (approved):
Thanks!

busterb (approved):

Issue #1015, opened 2024-03-12
3.8.3, 3.9.0 mingw-w64 x86_64 clang ASM CET builds crash on startup


vszakats:
Issue #907 returned with 3.8.3 and 3.9.0.

Building with CET enabled with mingw-w64 work fine with 3.8.2:

`CFLAGS=-fcf-protection=full`
`LDFLAGS=-Wl,-Xlink=-cetcompat`

With 3.8.3, it's crashing the same way as 3.8.1 did.

curl.exe exhibiting the issue:
https://curl.se/windows/dl-8.6.0_5/curl-8.6.0_5-win64-mingw.zip

vszakats:
What's happening is that `CFLAGS` makes the compiler define `__CET__`, which makes LibreSSL #include `cet.h`, which in turn sets the new macro `_CET_ENDBR` to `endbr32`, ending up with the same code that caused the crash in 3.8.1.

Changing `#if defined(__CET__)` to `#if defined(__CET__) && !defined(__MINGW32__)` fixes it.

Hrxn:
@vszakats 

Is that here the reason that curl `8.6.0_5` is not working at all for me (Win x64)?

vszakats:
@Hrxn: Yes, it's because of this issue.

busterb:
We should add this flag option to the CI tests as well. Thanks for the report.

I'm curious about one thing though; previous releases used a hard-coded [endbr64](https://github.com/libressl/openbsd/commit/c10c5b524e3121f42c9239d737dd7f975638f378), but it sounds like you're saying that mingw-w64 resolves  _CET_ENDBR to endbr32?

vszakats:
@busterb: Yes, that's what's happening:

t.S:
```c
#if defined(__CET__)
#include <cet.h>
#else
#define _CET_ENDBR
#endif

_CET_ENDBR
```

llvm-mingw 20240308 with LLVM 18.1.1
```terminal
$ x86_64-w64-mingw32-clang -E -fcf-protection=full t.S

# 1 "t.S"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 404 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "t.S" 2

# 1 "/path/to/llvm-mingw/lib/clang/18/include/cet.h" 1 3
# 44 "/path/to/llvm-mingw/lib/clang/18/include/cet.h" 3
[...]

PR #1032, opened 2024-03-26
mingw: re-enable endbr opcode, add workaround for llvm `cet.h` issue

vszakats:
- https://github.com/libressl/openbsd/pull/149 fixes to crash
  on startup issue, thus nuking all endbr opcodes is no longer
  necessary. This effectively enables CET in ASM code for
  MinGW x86_64 builds.
  A broader fix landed now via:
  https://github.com/openbsd/src/commit/db9cd57725789270c36763286ca8dc0a7c1b9d10

- llvm `cet.h` (as of v18) assigns `endbr32` instead of `endbr64`
  to `_CET_ENDBR` macro for MinGW x86_64 targets.
  https://github.com/llvm/llvm-project/blob/llvmorg-18.1.1/clang/lib/Headers/cet.h#L15-L35
  Work this around by mapping `endbr32` to `endbr64`.

Prerequisite: https://github.com/libressl/openbsd/pull/149
Fixes: https://github.com/libressl/portable/issues/1015

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -193,7 +193,7 @@ if(HOST_ASM_MINGW64_X86_64)
 		whrlpool/wp-mingw64-x86_64.S
 		cpuid-mingw64-x86_64.S
 	)
-	add_definitions(-Dendbr64=)
+	add_definitions(-Dendbr32=endbr64)
 	add_definitions(-DAES_ASM)
 	add_definitions(-DBSAES_ASM)
 	add_definitions(-DVPAES_ASM)
--- a/crypto/Makefile.am.mingw64-x86_64
+++ b/crypto/Makefile.am.mingw64-x86_64
@@ -21,6 +21,7 @@ ASM_X86_64_MINGW64 += cpuid-mingw64-x86_64.S
 EXTRA_DIST += $(ASM_X86_64_MINGW64)
 
 if HOST_ASM_MINGW64_X86_64
+libcrypto_la_CPPFLAGS += -Dendbr32=endbr64
 libcrypto_la_CPPFLAGS += -DAES_ASM
 libcrypto_la_CPPFLAGS += -DBSAES_ASM
 libcrypto_la_CPPFLAGS += -DVPAES_ASM


busterb:
Tthanks! We're going to need the change in libressl/openbsd#149 reviewed upstream in OpenBSD, which might take a bit. I might disable endbr* still temporarily while this works its way through, since we have a cert expiration update that's time-sensitive.

crypto/CMakeLists.txt
```diff
@@ -193,7 +193,7 @@ if(HOST_ASM_MINGW64_X86_64)
 		whrlpool/wp-mingw64-x86_64.S
 		cpuid-mingw64-x86_64.S
 	)
-	add_definitions(-Dendbr64=)
```
busterb, line 196: Don't forget crypto/Makefile.am.mingw64-x86_64 too

vszakats: Thanks for the reminder, I applied the same for autotools.

busterb (commented):
This looks fine to me. @botovq and friends, let us know how you want to handle the upstream change

botovq:
The removal of `_CET_ENDBR` will need more thinking. It needs to stay in `x86_64cpuid.pl` for at least OpenBSD.

We probably need to intercept `endbr64` in `x68_64-xlate.pl` and define it away if it appears in the `.init` section for MinGW, maybe around here.

https://github.com/openbsd/src/blob/master/lib/libcrypto/perlasm/x86_64-xlate.pl#L512

The current theory is that something is computing relocations on all opcodes, where it only expects `call`, leading to the crash.

vszakats:
> The removal of `_CET_ENDBR` will need more thinking. It needs to stay in `x86_64cpuid.pl` for at least OpenBSD.
> 
> We probably need to intercept `endbr64` in `x68_64-xlate.pl` and define it away if it appears in the `.init` section for MinGW, maybe around here.

MSVC CRT also expects a list a function pointers in its init section:
https://learn.microsoft.com/cpp/c-runtime-library/crt-initialization

botovq:
We will most likely remove more from this file which will solve your
problem. It will take a little bit of time. Thanks!

vszakats:
This will have to be rebased on the 3.9.1 tree, once it lands in this repo.

PR #1033, opened 2024-03-26
temporarily undefine endbr* for stable branch macos/windows builds

busterb:
Testing getting the next stable release out without endbr*/CET enabled in asm code for now while #1032 is considered for upstream. This should also fix #1023 in the release tarballs.

PR #1035, opened 2024-03-28
explicitly undef __CET__ on masm cpp

busterb:
Some systems (e.g. OpenBSD) unconditionally set __CET__ in the C preprocessor, which we don't want for masm yet. Eventually we need to figure out how to use the VS C preprocessor on assembly directly. This already went into the stable branches in #1033


busterb:
Note, this build is going to be broken for a while as upstream is making a lot of changes.

PR #1041, opened 2024-04-05
Add `rm libressl-*.tar.gz` to test scripts/test

nak3:
This patch adds `rm libressl-*.tar.gz`.

Currently some distribution (e.g. Fedora 39) `cd libressl-*` fails as:

```
+ cd libressl-3.9.0 libressl-3.9.0.tar.gz
./scripts/test: line 67: cd: too many arguments
```

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -27,6 +27,7 @@ if [ `uname` = "Darwin" ]; then
 
 	# test cmake
 	tar zxvf libressl-*.tar.gz
+	rm libressl-*.tar.gz
 	cd libressl-*
 
 	(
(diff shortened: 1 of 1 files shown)


nak3:
Other script has this `rm` as 
https://github.com/libressl/portable/blob/7362be429ccf7c8381cc3e7f91c1a57ff1a11f7f/.github/workflows/coverity.yml#L43-L45

Also, the issue does not happen on current MacOS 14 but does on Fedora 39. 
(The following `cd` command works on MacOS but Fedora (Linux) does not.)

```
$ mkdir foo1 foo2

$ cd foo*
-bash: cd: too many arguments
```

busterb:
What do you think about #1047 as an alternative solution?  I think being more explicit is more understandable.

nak3:
Yes, #1047 looks better! I close here in favor of #1047.

PR #1043, opened 2024-04-06
Modify CMake rules to only build uninstall if install is enabled

d3x0r:
Collapse the same if condition blocks into one block.  Include the uninstall rule when building install rules; if there was no install, shouldn't need uninstall (for this build; a previous build that was installed would have an uninstall rule).

There were several if( ENABLE_LIBRESSL_INSTALL ) in a row, hence the deleted endif()/if( ENABLE_LIBRESSL_INSTALL ) lines.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -509,9 +509,7 @@ if(ENABLE_LIBRESSL_INSTALL)
 		"${CMAKE_CURRENT_BINARY_DIR}/LibreSSLConfigVersion.cmake"
 		DESTINATION "${LIBRESSL_INSTALL_CMAKEDIR}"
 	)
-endif()
 
-if(ENABLE_LIBRESSL_INSTALL)
 	if(NOT MSVC)
 		# Create pkgconfig files.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -530,19 +528,17 @@ if(ENABLE_LIBRESSL_INSTALL)
...
+			"${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
+			IMMEDIATE @ONLY)
 
-	add_custom_target(uninstall
-		COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake)
-endif()
+		add_custom_target(uninstall
+			COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake)
+	endif()
+endif(ENABLE_LIBRESSL_INSTALL)
```
vszakats, line 544:
```suggestion
endif()
```
Suggesting this syntax tidy-up.

joshuasing (approved):

PR #1046, opened 2024-04-08
[Windows] Mark old_report_mode as unused

RandomInEqualities:
When compiling in release mode the _CrtSetReportMode macro gets replaced by void(0). This makes old_report_mode unused, which visual studio warns about.

### DIFF
--- a/crypto/compat/posix_win.c
+++ b/crypto/compat/posix_win.c
@@ -164,9 +164,10 @@ static void noop_handler(const wchar_t *expression,	const wchar_t *function,
 }
 
 #define BEGIN_SUPPRESS_IPH \
-	int old_report_mode = _CrtSetReportMode(_CRT_ASSERT, 0); \
-	_invalid_parameter_handler old_handler = _set_thread_local_invalid_parameter_handler(noop_handler)
+	const int old_report_mode = _CrtSetReportMode(_CRT_ASSERT, 0); \
+	const _invalid_parameter_handler old_handler = _set_thread_local_invalid_parameter_handler(noop_handler)
 #define END_SUPPRESS_IPH \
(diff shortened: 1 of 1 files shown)


busterb:
Could you add a comment in the code for reference later about why this is needed? Thanks!

crypto/compat/posix_win.c
```diff
@@ -165,6 +165,7 @@ static void noop_handler(const wchar_t *expression,	const wchar_t *function,
 
 #define BEGIN_SUPPRESS_IPH \
 	int old_report_mode = _CrtSetReportMode(_CRT_ASSERT, 0); \
+	(void)old_report_mode; \
 	_invalid_parameter_handler old_handler = _set_thread_local_invalid_parameter_handler(noop_handler)
 #define END_SUPPRESS_IPH \
 	_CrtSetReportMode(_CRT_ASSERT, old_report_mode); \
```
vszakats, line 171:
```suggestion
#define BEGIN_SUPPRESS_IPH \
	int old_report_mode = _CrtSetReportMode(_CRT_ASSERT, 0); \
	_invalid_parameter_handler old_handler = _set_thread_local_invalid_parameter_handler(noop_handler)
#define END_SUPPRESS_IPH \
	(void)old_report_mode; \
	_CrtSetReportMode(_CRT_ASSERT, old_report_mode); \
```
Wouldn't it be better to move the warning suppression into `END_SUPPRESS_IPH`? This way such warning would still correctly signal when missing an `END_SUPPRESS_IPH` by accident.

RandomInEqualities: Year I think that sounds good :)

RandomInEqualities:
I updated it and added const to the variables.

PR #1047, opened 2024-04-09
use a specific version number in test scripts

busterb:
This avoids issue reported in #1041 with file glob confusing `cd` in some cases.
Also remove extra newline from VERSION when update.sh runs.


nak3:
/lgtm

Thank you!

PR #1056, opened 2024-05-25
Remove timegm() compat

botovq:
Now that all uses of gmtime_r() and timegm() have been converted to OPENSSL_gmtime() and OPENSSL_timegm(), this is no longer needed.


joshuasing:
I think you may have already seen, however the `asn1time` test appears to be failing on linux/arm32.
Could this log output be related?
 ```
 asn1time.c: In function ‘asn1_gentime_test’:
asn1time.c:311:27: warning: passing argument 2 of ‘OPENSSL_timegm’ from incompatible pointer type [-Wincompatible-pointer-types]
  311 |  if (!OPENSSL_timegm(&tm, &a)) {
      |                           ^~
      |                           |
      |                           int64_t * {aka long long int *}
In file included from asn1time.c:20:
../include/openssl/posix_time.h:48:49: note: expected ‘time_t *’ {aka ‘long int *’} but argument is of type ‘int64_t *’ {aka ‘long long int *’}
   48 | int OPENSSL_timegm(const struct tm *tm, time_t *out);
      |                                         ~~~~~~~~^~~
```

**Comment by botovq** [comment]
Thanks. Yes, it's part of what's wrong here, I reverted that change I made.

But there's more. [The check][1] I tried to adapt also can't fail, as all the times in [asn1_gentime_tests][2] are long before the epoch wraps, so it makes no sense in the first place...

[1]: https://github.com/openbsd/src/blob/9251d8a3d62b18fbf1e0fcc83036d282fcf226bb/regress/lib/libcrypto/asn1/asn1time.c#L310-L311
[2]: https://github.com/openbsd/src/blob/9251d8a3d62b18fbf1e0fcc83036d282fcf226bb/regress/lib/libcrypto/asn1/asn1time.c#L93-L124

**Comment by joshuasing** [comment]
Ah, okay. That's interesting. I reran the failed CI, which seems to have pulled the "Revert previous" commit from libressl/openbsd, and it has passed.

**Review by joshuasing — APPROVED** [review]

**Comment by botovq** [comment]
It's not ready yet,  the asn1time now again has `timegm()` apparently none of our CI systems are missing it, though.

**Comment by botovq** [comment]
@joshuasing As you can see, arm32 is [completely busted][1] once you [add a test case][2] that tickles it.

```
FAIL: test 3 - times don't match, expected -2147483648 got -1
UTCTIME tests...
TIME tests...
ASN1_TIME_compare tests...
20161208193400Z vs. 20380119031408Z: want 1, got -1
20161208193400Z vs. -2147483648: want -1, got 1
20161208193400Z vs. -2147483648: want -1, got 1
19700101000000Z vs. 20380119031408Z: want 1, got -1
19700101000000Z vs. -2147483648: want -1, got 1
19700101000000Z vs. -2147483648: want -1, got 1
20150923032700Z vs. 20380119031408Z: want 1, got -1
20150923032700Z vs. -2147483648: want -1, got 1
20150923032700Z vs. -2147483648: want -1, got 1
20380119031408Z vs. 20161208193400Z: want -1, got 1
...
```

[1]: https://github.com/botovq/libressl-portable/actions/runs/9236559125/job/25412538486#step:3:1643
[2]: https://github.com/botovq/libressl-portable/blob/asn1time/patches/patch-asn1time.patch

Issue #1055, opened 2024-05-21
Unsupported platforms should fail earlier in the configuration process


CodingMarkus:
When trying to cross-compile, e.g. to web assembly, and I configure like 

```
emconfigure ./configure --host=wasm32-unknown-emscripten 
```

and then try to build it using

```
emmake make
```

The compilation fails with

```
  CC       libcrypto_la-cpt_err.lo
In file included from cpt_err.c:58:
In file included from /usr/local/include/openssl/opensslconf.h:15:
/usr/local/include/openssl/macros.h:147:4: error: "OPENSSL_API_COMPAT expresses an impossible API compatibility level"
  147 | #  error "OPENSSL_API_COMPAT expresses an impossible API compatibility level"
      |    ^
1 error generated.
```

But why on earth would the code even read a header file from the OpenSSL/LibreSSL library I currently have installed?  I am trying to build this stuff right now. My system should not even be required to have OpenSSL/LibreSSL installed, otherwise it creates a chicken-egg problem. IMHO, the source code should never reference any OpenSSL/LibreSSL header files in my current system or build environment, only from the source directory of the code I'm trying to build.

It even makes less sense since I'm cross compiling, so my locally installed headers must be totally irrelevant as my local system is not the target system the code will be running on.

If I remove that file, the build fails as well

```
/Library/Automount/Internal/Xcode/App/Xcode_15.3.app/Contents/Developer/usr/bin/make  all-am
  CC       libcrypto_la-cpt_err.lo
cpt_err.c:58:10: fatal error: 'openssl/opensslconf.h' file not found
   58 | #include <openssl/opensslconf.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

so you cannot cross-build LibreSSL unless you have either LibreSSL or OpenSSL installed on your build system?

CodingMarkus:
Okay, the problem is caused by `include/openssl/Makefile.am.tpl`. If the architecture is unknown or not explicitly handled, no `opensslconf.h` file is generated at all.

```
opensslconf.h: Makefile
	-echo "generating opensslconf.h ..."
if HOST_AARCH64
	-cp $(top_srcdir)/include/arch/aarch64/opensslconf.h opensslconf.h
endif
if HOST_ARM
	-cp $(top_srcdir)/include/arch/arm/opensslconf.h opensslconf.h
endif
if HOST_I386
[...]

PR #1061, opened 2024-05-28
Add Emscripten branch to arc4random

MoustaphaSaad:
The Emscripten platform already supports the necessary Linux API. The only missing component was adding a specific branch for Emscripten and including the existing Linux file.

This update enables all LibreSSL libraries to be built using the Emscripten toolchain.

---

**Emscripten Build Guide**

The only difference in the build process is that you need to provide the Emscripten toolchain file in the CMake configuration step. This can be done as follows:

```sh
mkdir build
cd build
cmake .. -DCMAKE_TOOLCHAIN_FILE=$EMSDK/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake -DENABLE_ASM=OFF -DLIBRESSL_TESTS=OFF
make
```

~Please note that not all test cases build successfully, which is why the `LIBRESSL_TESTS=OFF` flag is used. I have a preliminary draft for building and running the tests, achieving an initial 110/129 passing rate. Let me know if you would like another PR for that.~

---

**Update**

The tests are now fixed, so you can remove the `-DLIBRESSL_TESTS=OFF` flag from the guide above. You can run the tests using the usual `make test` command.

---

**Update regarding ASAN build**

Note: When enabling ASAN with Emscripten, the `ssl_get_shared_ciphers` test fails with the following error:
```
READ of size 4 at 0x124b53b4 thread T0
    #0 0x1a95f5 in strlcpy ../../../../../../../emsdk/emscripten/system/lib/libc/musl/src/string/strlcpy.c:24:33
    #1 0x1a93e7 in strlcat ../../../../../../../emsdk/emscripten/system/lib/libc/musl/src/string/strlcat.c:8:13
    #2 0x4f83 in main ../../ssl/ssl_lib.c:1735:7
    #3 0x800017da in Module._main /home/runner/work/portable/portable/build/tests/ssl_get_shared_ciphers.js:6106:102
```
This issue does not occur with regular Linux ASAN builds. It seems to be related to Emscripten's implementation of `strlcpy`. I found a similar issue [reported here](https://github.com/emscripten-core/emscripten/issues/7279#issuecomment-594103181), which was supposedly fixed 4 years ago. I'm not sure why this problem persists.


botovq:
Thanks. Before merging something like this, I'd like to know how well
this works.

> Please note that not all test cases build successfully, which is why
> the `LIBRESSL_TESTS=OFF` flag is used. I have a preliminary draft for
> building and running the tests, achieving an initial 110/129 passing
> rate. Let me know if you would like another PR for that.

What kinds of changes are required and what kinds of failures do occur?
How much effort do you estimate will it be to get them all to pass?

MoustaphaSaad:
The first issue I encountered was related to test execution. In WebAssembly, you need to pass the generated `a.out.js` to `node` to run it, as it's not an executable by itself. To address this, I introduced a function like the following:

```cmake
function(add_platform_test TEST_NAME)
    if (EMSCRIPTEN)
        add_test(NAME ${TEST_NAME} COMMAND node ${ARGN})
    else()
        add_test(NAME ${TEST_NAME} COMMAND ${ARGN})
    endif()
endfunction()
```

This adjustment resolves most of the issues, and as mentioned, the majority of the test cases pass out of the box.

The remaining failures are mainly due to the lack of direct file system access in WebAssembly. These can be mitigated by using `--preload-file` link flags, as shown below:

```cmake
set_target_properties(aeadtest PROPERTIES LINK_FLAGS
    "--preload-file ${TEST_SOURCE_DIR}/aeadtests.txt@${TEST_SOURCE_DIR}/aeadtests.txt"
)
```

The third category of failures stems from missing feature support in Emscripten, such as the lack of `sigsuspend` support (used in the `explicit_bzero` test) and `fork` support (used in the `arc4randomforktest`). I generally disable these tests for Emscripten.

I have not yet investigated or fixed all the failures, with approximately 10 tests still failing. If these changes and fixes are acceptable, I can continue working on the remaining test cases and submit them either in a separate PR or within this same PR.

I think the effort to fix the remaining issues isn't big; it can be investigated and fixed in a day or two. @botovq Please let me know your preference.

botovq:
```cmake
function(add_platform_test TEST_NAME)
    if (EMSCRIPTEN)
        add_test(NAME ${TEST_NAME} COMMAND node ${ARGN})
    else()
        add_test(NAME ${TEST_NAME} COMMAND ${ARGN})
    endif()
endfunction()
```

Makes sense, I'd be happy to use that abstraction.

```cmake
set_target_properties(aeadtest PROPERTIES LINK_FLAGS
    "--preload-file ${TEST_SOURCE_DIR}/aeadtests.txt@${TEST_SOURCE_DIR}/aeadtests.txt"
)
```

This seems sensible as well.

I think disabling the `explicit_bzero()` and `arc4random()` tests in a first pass makes sense as well.

I am not familiar with emscripten but if it is as little effort as you make it out to be, I'd be happy to add a few patches and abstractions for preliminary support of it. Surely there are some possibilities to test this in CI?

Instead of landing it piecemeal, I think it would be preferable to add CI for emscripten to this PR and add the patches you have on top. Then we can see how far away we are and can worry about how to merge it once we get close.

tests/aeadtest.sh
```diff
@@ -3,6 +3,8 @@ set -e
 TEST=./aeadtest
 if [ -e ./aeadtest.exe ]; then
 	TEST=./aeadtest.exe
+elif [ -e ./aeadtest.js ]; then
```
busterb, line 6: don't forget to add/push aeadtest.js as well

MoustaphaSaad: I don't think it should included.
aeadtest.js file is generated by emscripten when it compiles the aeadtest target to wasm

MoustaphaSaad:
Hello @botovq,

As agreed, I've updated the test cases and fixed nearly all of them except for the following:

- **explicit_bzero**: This relies on `sigsuspend`, which is not supported in Emscripten.
- **arc4randomforktest**: This relies on `fork`, which is also not supported in Emscripten.
- **testdsa, testenc, and testrsa**: These don't have CMake targets (executables) and instead use OpenSSL, which accesses many input and output files. This makes adding a `--preload-file` flag infeasible, so I decided to disable them.

Other than these five, all remaining test cases (125 in total) pass.

I've also written an initial CI script to test Emscripten builds, but I haven't tested it because I can't run or trigger GitHub Actions CI.

I've structured my edits into atomic commits, with each commit fixing a single test case. This should make reviewing the entire PR easier by going through the series of commits.

Looking forward to your feedback.

tests/CMakeLists.txt
```diff
@@ -252,6 +252,7 @@ add_platform_test(bn_to_string bn_to_string)
 # bn_unit
 add_executable(bn_unit bn_unit.c)
 target_link_libraries(bn_unit ${OPENSSL_TEST_LIBS})
+set_target_properties(bn_unit PROPERTIES LINK_FLAGS "-sALLOW_MEMORY_GROWTH")
 add_platform_test(bn_unit bn_unit)
```
botovq, line 256: This should probably only be set for emscripten.

botovq:
Thanks. Splitting the commits this way is very hepful. This looks pretty good at a first glance and the fallout from the `-sALLOW_MEMORY_GROWTH` flag should be trivial to fix I will review in more detail in the coming days.

MoustaphaSaad:
Awesome! I've fixed the `-sALLOW_MEMORY_GROWTH` link flag. Looking forward to hearing from you.

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: mymindstorm/setup-emsdk@v14
+        with:
+          version: 3.1.60
+
+      - name: "Prepare repository"
+        run: "./autogen.sh"
+
+      - name: "Configure CMake"
+        run: cmake -Bbuild -DCMAKE_CROSSCOMPILING_EMULATOR=$EMSDK/node/18.20.3_64bit/bin/node -DCMAKE_TOOLCHAIN_FILE=$EMSDK/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake -DENABLE_ASM=OFF
+
```
busterb, line 35: This seems to run fine locally with the latest emsdk. Is it more typical to build as you've done in the github actions by setting the toolchain parameters manually, or use wrappers like emcmake / emconfigure like below? It doesn't seem to be necessary to manually disable ASM either - we probably shouldn't require a switch if this is the proper default for the platform.

```
$ emcmake cmake .. -GNinja
configure: cmake .. -GNinja -DCMAKE_TOOLCHAIN_FILE=/home/bcook/projects/emsdk/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake -DCMAKE_CROSSCOMPILING_EMULATOR=/home/bcook/projects/emsdk/node/18.20.3_64bit/bin/node
-- The ASM compiler identification is unknown
-- Found assembler: /home/bcook/projects/emsdk/upstream/emscripten/emcc
-- Warning: Did not find file Compiler/-ASM
-- Setting build type to 'Release' as none was specified.
...
```

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,68 @@
...
+    permissions:
+      contents: read
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v4
+
+      - name: "Setup emsdk"
+        uses: mymindstorm/setup-emsdk@v14
+        with:
+          version: 3.1.60
```
busterb, line 28: Using the latest emscripten package that is bundled with Debian 'trixie', this fails at build time like follows:

```
/home/bcook/projects/libressl/portable/crypto/cryptlib.c:373:21: error: variable has incomplete type 'struct syslog_data'
        struct syslog_data sdata = SYSLOG_DATA_INIT;
                           ^
/home/bcook/projects/libressl/portable/crypto/cryptlib.c:373:9: note: forward declaration of 'struct syslog_data'
        struct syslog_data sdata = SYSLOG_DATA_INIT;
               ^
/home/bcook/projects/libressl/portable/crypto/cryptlib.c:377:2: error: call to undeclared function 'vsyslog_r'; ISO C99 and later do not support implicit function declarations [-Werror,-Wimplicit-function-declaration]
        vsyslog_r(LOG_CONS|LOG_LOCAL2, &sdata, fmta, ap);
        ^
/home/bcook/projects/libressl/portable/crypto/cryptlib.c:377:2: note: did you mean 'vsyslog'?
/usr/share/emscripten/cache/sysroot/include/syslog.h:68:6: note: 'vsyslog' declared here
void vsyslog (int, const char *, va_list);
     ^
/home/bcook/projects/libressl/portable/crypto/cryptlib.c:386:16: error: call to undeclared function 'getprogname'; ISO C99 and later do not support implicit function declarations [-Werror,-Wimplicit-function-declaration]
            getuid(), getprogname(), file, line, assertion);
                      ^
/home/bcook/projects/libressl/portable/crypto/cryptlib.c:373:29: error: use of undeclared identifier 'SYSLOG_DATA_INIT'
        struct syslog_data sdata = SYSLOG_DATA_INIT;
                                   ^
4 errors generated.
```

Looks like it is an older version. Is it worth supporting, or should there be checks on version to make sure this fails earlier: 

```
$ dpkg -l |grep emscripten
ii  emscripten                                     3.1.6~dfsg-7                           all          LLVM-to-JavaScript Compiler
```

Hopefully #1055 isn't a generalization of the kinds of issues we might receive from the Emscripten community if this lands and doesn't work out of the box, or fail very explicitly, for everyone's environment.

.github/workflows/emscripten.yml
MoustaphaSaad: Good point, using emcmake is the recommended way in Emscripten website, I'll change it to use it instead of feeding it the toolchain file manually

.github/workflows/emscripten.yml
MoustaphaSaad: After conducting some research, I've determined that the oldest version this PR supports is 3.1.44 (released on 21, July, 2023). According to the Emscripten website, they recommend installing it via Git, so supporting older versions doesn't seem necessary as it's not the recommended method.

I can implement a version check using CMake. While it might not be the cleanest approach, it should work. This method involves invoking `emcc --version` and using regex to parse the version number. Here's a rough outline of the solution. Should I proceed with this approach? What do you think?

```cmake
# Execute the 'emcc --version' command and capture the output
execute_process(
    COMMAND emcc --version
    OUTPUT_VARIABLE EMSCRIPTEN_VERSION_OUTPUT
    OUTPUT_STRIP_TRAILING_WHITESPACE
)

# Parse the version string from the output
string(REGEX MATCH "[0-9]+\\.[0-9]+\\.[0-9]+" EMSCRIPTEN_VERSION ${EMSCRIPTEN_VERSION_OUTPUT})
```

busterb: I'm inclined to not worry too much about it, and see if there are any reports in the wild. But we may want to add a check later if this becomes a common foot-gun of using ancient versions packaged with distributions.

MoustaphaSaad: I feel the same way

botovq (commented):
I was wondering if the setting of the link flags warrants factoring into a helper, maybe `add_platform_test_with_fs_access()` or similar..

Generally I would be happy about as few `if(EMSCRIPTEN)` stanzas as possible.

tests/ssltest.sh
```diff
@@ -4,17 +4,23 @@ set -e
 ssltest_bin=./ssltest
 if [ -e ./ssltest.exe ]; then
 	ssltest_bin=./ssltest.exe
+elif [ -e ./ssltest.js ]; then
+	ssltest_bin="node ./ssltest.js"
```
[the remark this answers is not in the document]
botovq: I think they are result from compilation, just like on windows you get an .exe: https://github.com/libressl/portable/pull/1061#discussion_r1622812581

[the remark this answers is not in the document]
joshuasing: Ah okay, thanks!

tests/CMakeLists.txt
```diff
@@ -22,32 +22,40 @@ add_definitions(-D_PATH_SSL_CA_FILE=\"${CMAKE_CURRENT_SOURCE_DIR}/../cert.pem\")
...
+	else()
+		add_test(NAME ${TEST_NAME} COMMAND ${ARGN})
+	endif()
+endfunction()
+
 # aeadtest
 add_executable(aeadtest aeadtest.c)
 target_link_libraries(aeadtest ${OPENSSL_TEST_LIBS})
 if(NOT WIN32)
 	add_test(NAME aeadtest COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/aeadtest.sh)
```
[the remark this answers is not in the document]
MoustaphaSaad: We have two types of tests: those launched directly and those launched via a script. For the tests launched directly, I use `add_platform_test`, which runs the test through Node.js. For the tests launched by a script, I don't use `add_platform_test`. Instead, I modify the script to launch the executable using Node.js. This change is demonstrated in `aeadtest.sh` with the following snippet:

```sh
elif [ -e ./aeadtest.js ]; then
	TEST="node ./aeadtest.js"
fi
```

tests/CMakeLists.txt
```diff
@@ -402,7 +402,8 @@ add_test(exdata_test exdata_test)
 
 # explicit_bzero
 # explicit_bzero relies on SA_ONSTACK, which is unavailable on Windows
-if(NOT WIN32)
+# explicit_bzero relies on sigsuspend, which is unavailable on Emscripten
```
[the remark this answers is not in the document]
MoustaphaSaad: sure thing

MoustaphaSaad:
> I was wondering if the setting of the link flags warrants factoring into a helper, maybe `add_platform_test_with_fs_access()` or similar..
> 
> Generally I would be happy about as few `if(EMSCRIPTEN)` stanzas as possible.

I was considering that as well. I also agree that it would be simpler with fewer if(EMSCRIPTEN) statements.

MoustaphaSaad:
@botovq 
The issue with moving the `--preload-file` linking flag to the `add_platform_test` function is that some tests are invoked through a `.sh` script. In these cases, we don't call `add_platform_test`; instead, we use `add_test` for the reasons explained in [this comment](https://github.com/libressl/portable/pull/1061#discussion_r1631172438). Therefore, for these tests, we'll need to keep the `--preload-file` linker flag. Given that, I suggest we keep things as they are now, with tests that require file system access having the `--preload-file` flags passed individually.

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,68 @@
...
+        run: emcmake cmake -Bbuild
+
+      - name: "Build"
+        run: cmake --build build --config Release
+
+      - name: "Test"
+        run: ctest --test-dir build -C Release --output-on-failure
+
+  # Test ASAN with and without ASM enabled.
+  test-asan:
```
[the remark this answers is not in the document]
MoustaphaSaad: good point, sorry, I just copied the test job and forgot to set the env flags, will fix that

.github/workflows/emscripten.yml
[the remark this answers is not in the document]
MoustaphaSaad: sure thing, will fix

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: "Checkout repository"
+        uses: actions/checkout@v4
+
+      - name: "Setup emsdk"
+        uses: mymindstorm/setup-emsdk@v14
+        with:
+          version: 3.1.60
+
+      - name: "Prepare repository"
+        run: "./autogen.sh"
```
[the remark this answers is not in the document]
MoustaphaSaad: sure thing, will fix

botovq:
I see. Thanks for the additional explanations. If it's not easily feasible, that's fine with me. I don't think we'll add many more types of tests anytime soon, so I should be able to navigate these cases when adding new ones.

As far as I'm concerned, I'm fine with merging this with whatever tweaks @busterb and @joshuasing would still want to see on top. I think it's best to deal with it in tree and see how it goes.

If it doesn't get in the way too badly (I don't currently see how it would) until the release (october), I think we should ship it as experimental for a cycle or two.

MoustaphaSaad:
@botovq
As I was adding the missing `-fsanitize=address` flags as requested in [this comment](https://github.com/libressl/portable/pull/1061#discussion_r1629350179), some of the tests started failing due to an Out Of Memory (OOM) error. According to [the Emscripten website](https://emscripten.org/docs/debugging/Sanitizers.html#address-sanitizer), it's recommended to use `-sALLOW_MEMORY_GROWTH` when address sanitizer is used, as it requires more memory than the default 16 MB.

To simplify the handling of Emscripten linking flags, I propose adding a `prepare_emscripten_test_target` function to include the `--preload-file` and `-sALLOW_MEMORY_GROWTH` flags. We can invoke this function for tests that need it.

The function would look like this:

```cmake
function(prepare_emscripten_test_target TARGET_NAME)
	if(EMSCRIPTEN)
		set_target_properties(${TARGET_NAME} PROPERTIES LINK_FLAGS
			"--preload-file ${TEST_SOURCE_DIR} -sALLOW_MEMORY_GROWTH")
	endif()
endfunction()
```

What do you think? Should I go with this simplification instead of adding more `if (EMSCRIPTEN)` branches?

botovq:
I like it.

tests/CMakeLists.txt
```diff
@@ -705,46 +745,60 @@ endif()
 set_tests_properties(ssltest PROPERTIES ENVIRONMENT "srcdir=${TEST_SOURCE_DIR}")
 
 # testdsa
-if(NOT MSVC)
-	add_test(NAME testdsa COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/testdsa.sh)
-else()
-	add_test(NAME testdsa COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/testdsa.bat $<TARGET_FILE:openssl>)
+# testdsa doesn't have an executable instead it uses openssl executable which access various files for IO.
```
botovq, line 748: I think we should group these three tests into a single `if(NOT EMSCRIPTEN)` block then we need this comment only once:

```
# testdsa, testenc and testrsa rely on the openssl executable which accesses various files for IO
# adding these files to --preload-file is infeasible, so we skip these tests.
```

botovq (changes requested):
I'm still fine with this modulo one small request.

tests/CMakeLists.txt
MoustaphaSaad: should be resolved now

MoustaphaSaad:
Note: When enabling ASAN with Emscripten, the `ssl_get_shared_ciphers` test fails with the following error:
```
READ of size 4 at 0x124b53b4 thread T0
    #0 0x1a95f5 in strlcpy ../../../../../../../emsdk/emscripten/system/lib/libc/musl/src/string/strlcpy.c:24:33
    #1 0x1a93e7 in strlcat ../../../../../../../emsdk/emscripten/system/lib/libc/musl/src/string/strlcat.c:8:13
    #2 0x4f83 in main ../../ssl/ssl_lib.c:1735:7
    #3 0x800017da in Module._main /home/runner/work/portable/portable/build/tests/ssl_get_shared_ciphers.js:6106:102
```
This issue does not occur with regular Linux ASAN builds. It seems to be related to Emscripten's implementation of `strlcpy`. I found a similar issue [reported here](https://github.com/emscripten-core/emscripten/issues/7279#issuecomment-594103181), which was supposedly fixed 4 years ago. I'm not sure why this problem persists.

botovq:
I don't know how to fix that. That string is very unlikely to overflow the buffer as it is much larger than the concatenation of all ciphers we have.

Maybe it's worth a try to disable the strlcpy/strlcat detection and force use of our compat implementations for Emscripten?

https://github.com/libressl/portable/blob/46ddd6b184d211b56d8813069f62cb5e0ba32a1d/CMakeLists.txt#L223-L231

.github/workflows/emscripten.yml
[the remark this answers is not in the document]
joshuasing: Thanks!

MoustaphaSaad:
@botovq Awesome, I tested that on my machine and that indeed fixed the ASAN errors, Thanks for the nice idea

botovq:
Nice, thank you - all green now.

I'd say this is an acceptable workaround for the time being. The actual compat symbols used are prefixed with `libressl_`,  so this will not conflict if someone should link in another C module using them.

It would be nice to root cause this at some point and see if upstream (or perhaps musl?) could fix this. I don't think we're doing anything out of the ordinary in this test and if Emscripten's versions break with that, someone's made a bad mistake (either us, Emscripten or musl).

Pending further requests or okays from @busterb and @joshuasing I think we can merge this soon.

busterb (approved):

joshuasing (approved):

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,71 @@
+# GitHub Actions workflow to run tests on Linux.
```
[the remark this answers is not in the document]
MoustaphaSaad: fixed, thanks

botovq:
@MoustaphaSaad many thanks for your work, I have merged this now.

I have kept the ChangeLog entry very minimal for now. Feel free to make
a PR adding a couple of lines there, maybe adding a few hints to the
README.md with one or two hints on how to use this.

botovq:
Somehow github didn't recognize the merge as it usually does. The code is now in the master branch.

MoustaphaSaad:
> @MoustaphaSaad many thanks for your work, I have merged this now. I have kept the ChangeLog entry very minimal for now. Feel free to make a PR adding a couple of lines there, maybe adding a few hints to the README.md with one or two hints on how to use this.

Awesome, Thank you
Will surely add Emscripten entries to the README.md file

PR #1063, opened 2024-05-29
ci: add event number to concurrency group

joshuasing:
Add `github.event.number` to concurrency groups in CI workflows.
This should make sure that pushes to branches have separate groups than pull requests, and should prevent any conflicts that may cause unexpected skips/cancellations.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -4,7 +4,7 @@ name: "Android"
 on: [push, pull_request]
 
 concurrency:
-  group: "${{ github.workflow }}-${{ github.ref }}"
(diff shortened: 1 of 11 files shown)


botovq:
I can't judge this, but if you think this makes sense independently of the weird github-runner breakage, I'll happily merge, once github has sorted out their CI.

botovq:
So... close or merge? :)

joshuasing:
> So... close or merge? :)

I think merging this is a good idea, it may prevent future collisions, especially if someone creates a pull request from a branch in this repository.

PR #1067, opened 2024-06-20
Add Emscripten Compatibility and Configuration Note

MoustaphaSaad:
This PR adds Emscripten in the list of compatible operating systems. Additionally, it provides a note on using `emcmake` for configuring LibreSSL with Emscripten.


botovq:
Thanks!

PR #1070, opened 2024-07-13
include/CMakeLists.txt: error out when architecture is not supported

tpetazzoni:
Since commit b98c56fd5e86fa76eb55a98a2266a923f64f705c ("stop exposing hidden defines in public opensslconf.h"), only a subset of CPU architectures is supported by libressl. However, when a build is attempted for an unsupported architecture, it doesn't fail with a clear error message at configure time, but it fails with a weird error at build time:

crypto/../include/openssl/rc4.h:75:9: error: unknown type name 'RC4_INT'

In order to help users, let's bail out nicely at configure time with a good error message when the CPU architecture is not supported.


joshuasing (approved):

botovq:
This matches part of #1060. Thanks!

Issue #1069, opened 2024-07-09
posix_win.c: is_socket(int fd) fails at any socket whose value also exists as a file descriptor


datadiode:
In such situation, v3.8.1+ assumes that fd is not a socket but a file descriptor,
while earlier versions always assumed fd to be a socket in first place.

In other words, v3.8.0 and earlier gave a socket precedence over a file descriptor with the same value,
while v3.8.1 and later give a file descriptor precedence over a socket with the same value.

The discussion at #266 already outlines possible fixes.

Code snippet to likely produce a socket vs. file descriptor conflict:
```C++
#include <io.h>
#include <stdio.h>
#include <WinSock2.h>

int main(int argc, const char *argv[])
{
    WSADATA wsaData;
    WSAStartup(MAKEWORD(2, 2), &wsaData);

    int sd = (int)socket(AF_INET, SOCK_STREAM, 0);

    // Use up file descriptors until fopen() fails
    int fd = -1;
    while (FILE* f = fopen(argv[0], "rb"))
        fd = _fileno(f);

    intptr_t sh = _get_osfhandle(sd);
    intptr_t fh = _get_osfhandle(fd);

    printf("The operating-system file handle associated with fd is %p\n", fh);
    printf("The operating-system file handle associated with sd is %p\n", sh);

    return 0;
}
```
Employing _dup2() to achieve the same:
```C++
#include <io.h>
#include <stdio.h>
#include <WinSock2.h>

int main(int argc, const char *argv[])
{
    WSADATA wsaData;
    WSAStartup(MAKEWORD(2, 2), &wsaData);

    int sd = (int)socket(AF_INET, SOCK_STREAM, 0);

    // Duplicate stdout descriptor to produce a collision with the socket
    _dup2(1, sd);

    intptr_t sh = _get_osfhandle(sd);

    printf("The operating-system file handle associated with sd is %p\n", sh);

    return 0;
}
```

Ono-Sendai:
Urghh that's horrible.
LibreSSL should really use my solution I posted here: https://github.com/libressl/portable/issues/266#issuecomment-2227664358

busterb:
Closing this out, thanks for the help @datadiode and @Ono-Sendai

PR #1071, opened 2024-07-17
Fix for #1069

datadiode:
The proposed change makes file descriptors created through posix_open() distinguishable from sockets by having them take odd values only.

### DIFF
--- a/crypto/compat/posix_win.c
+++ b/crypto/compat/posix_win.c
@@ -41,6 +41,16 @@ posix_fopen(const char *path, const char *mode)
 	return fopen(path, mode);
 }
 
+static int
+oddify_fd(int fd)
+{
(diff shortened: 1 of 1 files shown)


busterb:
Hi @datadiode, interesting solution. How does this work when the file descriptors are passed in by the calling program or library though?

datadiode:
You mean if they mean to pass a file descriptor to e.g. SSL_set_fd()?
I never needed to, but I guess they could oddify it in the same way on their side to produce a file descriptor which LibreSSL can distinguish from a socket.

busterb:
I finally got a chance to test this out. It seems to break console IO when using openssl.exe. As soon as there is console IO, the program exits:
image

PR #1075, opened 2024-08-02
Fix MinGW-w64 assembler check in CMake always failing

rfl890:
On windows, CMAKE_SYSTEM_PROCESSOR will never be "x86_64", only "AMD64". This is accounted for by setting the HOST_X86_64 variable, but for some reason it's not used at all. This fixes that issue.


botovq:
Thanks. This looks right to me. The CI failure is unrelated and was fixed in HEAD. Could you please rebase and push again to see if it's all good now?

rfl890:
It seems https://github.com/libressl/openbsd/commit/3d1118fa05274d1f3789ca743e76aba59045acf7 is the offending commit here

botovq:
Yeah, that or the previous one. looking.

botovq:
Thanks. Apologies for these hiccups.

PR #1078, opened 2024-08-02
Attempt to fix 2038 problem with MSVC

### LINKED ISSUE #1076: gettimeofday implementation for MSVC cannot handle dates beyond 2038
MSVC gettimeofday implementation rely on the definition of timeval provided by winsock2.h
https://learn.microsoft.com/en-us/windows/win32/api/winsock2/ns-winsock2-timeval

typedef struct timeval {
  long tv_sec;
  long tv_usec;
} TIMEVAL, *PTIMEVAL, *LPTIMEVAL;

this give us potential errors when dealing with dates bigger than the year 2038, i.e. OpenSSH-Portable hangs if the machine date is set to a year bigger than 2039

[Connection is stuck when time on SSH Server is set after the Year-2038 Problem](https://github.com/PowerShell/Win32-OpenSSH/issues/2233)

botovq:
This might fix #1076

### DIFF
--- a/apps/openssl/compat/poll_win.c
+++ b/apps/openssl/compat/poll_win.c
@@ -248,9 +248,9 @@ poll(struct pollfd *pfds, nfds_t nfds, int timeout_ms)
 	timespent_ms = 0;
 	wait_rc = WAIT_FAILED;
 
-	if (timeout_ms < 0)
+	looptime_ms = (timeout_ms > 100 || timeout_ms == -1) ? 100 : timeout_ms;
+	if (timeout_ms == -1)
 		timeout_ms = INFINITE;
-	looptime_ms = timeout_ms > 100 ? 100 : timeout_ms;
 
 	do {
 		struct timeval tv;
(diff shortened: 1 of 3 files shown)


datadiode:
You may want to wrap winsock select() as well to ensure that it will still see the timeval struct layout it expects.

LainOTN2:
Tested on OpenSSH portable, and it fixes the issues on Win32 builds.

busterb:
I made a change to poll_win.c to ensure that we use the appropriate timeval definition with select with Windows. We were anyway, because sys/time.h isn't used in Windows' select definition, but still just in case.

This also fixes an issue where select was given {0, 4294967296} when poll was given a timeout value of -1, which caused poll() to spin with no timeout internally, noticed while logging internal behavior with `openssl s_client`.

busterb:
Going to land this now

botovq:
On Mon, Oct 07, 2024 at 08:22:28PM -0700, Brent Cook wrote:
> Going to land this now

Thanks. Sorry, didn't get around to reviewing it yesterday. Looks good
to me.

PR #1079, opened 2024-08-02
Switch to check_symbol_exists() in simplest cases

### LINKED ISSUE #1077: cmake: prefer check_symbol_exists() over check_function_exists()
The Android CI started failing because `check_function_exists()` detected that `pipe2` was available, but then it errored out because `unistd.h` doesn't contain a prototype:

```
/home/runner/work/portable/portable/tests/tlstest.c:244:6: error: call to undeclared function 'pipe2'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  244 |         if (pipe2(cfds, O_NONBLOCK) == -1)
      |             ^
```

The [check_function_exists()][1] documentation calls out this trap.

> **Note**
>
> Prefer using [CheckSymbolExists](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html#module:CheckSymbolExists) instead of this module, for the following reasons:
>
>    check_function_exists() can't detect functions that are inlined in headers or specified as a macro.
>
>    check_function_exists() can't detect anything in the 32-bit versions of the Win32 API, because of a mismatch in calling conventions.
>
>    check_function_exists() only verifies linking, it does not verify that the function is declared in system headers.

I pushed [a fix][2] for the specific issue with `pipe2()`, but I suspect that we should change all function detections to `check_symbol_exists()`.

[1]: https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html
[2]: https://github.com/libressl/portable/commit/161984f56f2953ebedc1a97056e4fd9cb938d672

botovq:
There are a few exceptions. Some require _GNU_SOURCE, some are in the weird strings.h header, some are probably too new to be found in the standard locations.

Fixes #1077


CMakeLists.txt
```diff
@@ -200,40 +200,42 @@ else()
...
 endif()
 
+# XXX - needs _GNU_SOURCE on linux
 check_function_exists(asprintf HAVE_ASPRINTF)
 if(HAVE_ASPRINTF)
 	add_definitions(-DHAVE_ASPRINTF)
 endif()
 
-check_function_exists(getopt HAVE_GETOPT)
+check_symbol_exists(getopt "unistd.h" iHAVE_GETOPT)
```
ikspress, line 209: A typo?

busterb (approved):
No problems with this that I could see.

PR #1084, opened 2024-08-21
Append version to dll on CMake

### LINKED ISSUE #1083: Inconsistent DLL name built by CMake and autotools
In CMakeLists.txt, it forces append major version to the dynamic libraries, but it doesn't work as expected, xxxx_POSTFIX is empty.
For example, CRYPTO_POSTFIX is empty, so CMake generates `libcrypto.dll` instead of `libcrypto-53.dll`.
https://github.com/libressl/portable/blob/3b545a08d6c14d67b6d1bead1e963391571b5182/crypto/CMakeLists.txt#L863-L872
Additionally, it's incorrect that forces append version to `OUTPUT_NAME` in CMake. When `CMAKE_DLL_NAME_WITH_SOVERSION` enabled and CRYPTO_POSTFIX is correct, CMake will generates `libcrypto-53-53.dll`.
**Solution:**
Just remove `CRYPTO_POSTFIX` `SSL_POSTFIX` and `TLS_POSTFIX` variables and let users to control it with `CMAKE_DLL_NAME_WITH_SOVERSION`.

ikspress:
Fixes #1083

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -50,7 +50,7 @@ jobs:
 
       - name: "Configure CMake"
         shell: cmd
-        run: cmake -Bbuild -G "${{ matrix.generator }}" -A ${{ matrix.arch }} DBUILD_SHARED_LIBS=${{ matrix.shared }} -DCMAKE_INSTALL_PREFIX=../local
+        run: cmake -Bbuild -G "${{ matrix.generator }}" -A ${{ matrix.arch }} -D BUILD_SHARED_LIBS=${{ matrix.shared }} -D CMAKE_INSTALL_PREFIX=../local
 
       - name: "Build"
         shell: cmd
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.16.4)
+cmake_minimum_required (VERSION 3.27)
 if(MSVC)
 	cmake_policy(SET CMP0091 NEW)
 endif()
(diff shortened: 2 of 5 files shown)


botovq (approved):
Thanks, this makes sense to me. I'm fine with doing this, but I would like to give  @busterb some time to comment.

busterb:
This seems like the right way to do this. Thanks @ikspress 

This does bump our minimum CMake version to 3.27, which I'm fine with https://cmake.org/cmake/help/latest/prop_tgt/DLL_NAME_WITH_SOVERSION.html

To get the unversioned behavior, a user can set CMAKE_PLATFORM_NO_VERSIONED_SONAME

busterb (approved):
The build failures in the CI re-run are just from this needing a rebase for changes in master openbsd branch since it was submitted.

vszakats:
> This seems like the right way to do this. Thanks @ikspress
> 
> This does bump our minimum CMake version to 3.27, which I'm fine with https://cmake.org/cmake/help/latest/prop_tgt/DLL_NAME_WITH_SOVERSION.html
> 
> To get the unversioned behavior, a user can set CMAKE_PLATFORM_NO_VERSIONED_SONAME

This CMake version is so new (2023-07-17), that this new requirement breaks
building LibreSSL on Debian stable, which [has](https://packages.debian.org/search?suite=bookworm&searchon=names&keywords=cmake) 3.25.1 when writing this.

Ref: https://github.com/curl/curl-for-win/actions/runs/11079065050/job/30787458202#step:3:12090

Anything below that would be nice, but I find that even 3.20.0 is too
new for some cases. Debian Bullseye [has](https://packages.debian.org/search?suite=bullseye&searchon=names&keywords=cmake) 3.18.4.

PR #1085, opened 2024-08-21
Fix includedir on CMake

ikspress:
Remove the hardcoded `include`.


botovq (commented):
I have no objection to doing this. Out of curiosity, is there an environment where this matters?

ikspress:
> I have no objection to doing this. Out of curiosity, is there an environment where this matters?

Yes, I'm building LibreSSL for myself. I configured it with `--DCMAKE_INSTALL_INCLUDEDIR=include/libressl` `-DCMAKE_INSTALL_LIBDIR=lib/libressl` to resolve naming conflicts with system's OpenSSL.

Now, libcrypto.pc works properly.

```
#libcrypto pkg-config source file

prefix=/mingw64
exec_prefix=${prefix}
libdir=${exec_prefix}/lib/libressl
includedir=${prefix}/include/libressl

Name: LibreSSL-libcrypto
Description: LibreSSL cryptography library
Version: 3.9.2
Libs: -L${libdir} -lcrypto
Libs.private:   -lws2_32 -lntdll -lbcrypt
Cflags: -I${includedir}
```

PR #1090, opened 2024-08-25
ci: update actions

joshuasing:
Update a few GitHub Actions to the latest versions. This includes some updates that change the version of Node used to run the actions, as Node.js 16 is EOL (https://nodejs.org/en/blog/announcements/nodejs16-eol, https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/).

Additionally, use the `softprops/action-gh-release` action for uploading Windows release artifacts instead of [`actions/upload-release-asset`](https://github.com/actions/upload-release-asset) as it is deprecated.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -26,7 +26,7 @@ jobs:
         dry-run: false
         language: c++
     - name: Upload Crash
-      uses: actions/upload-artifact@v3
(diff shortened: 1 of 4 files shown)


joshuasing:
@botovq If you see any more deprecation (or other) warnings in GitHub Actions, let me know. I think this should resolve most of them for now. Thanks :)

PR #1104, opened 2024-10-08
don't override language property for ASM files as C

busterb:
As of CMake 3.20, this causes the files to be built _as_ C, instead of just with a C compiler. This also properly specifies the languages in the project call.

This should fix issues that came from bumping CMake min version as in #1100


botovq (approved):
This looks good to me. Please merge if green.

PR #1108, opened 2024-10-15
Fix CMake build on FreeBSD

botovq:
FreeBSD's libc has a stub implementation of pthread_once() that returns ENOSYS and doesn't seem to call the init routine. You need to link with pthread for this to work.  This PR does this and fixes regress failures for CMake builds on this platform.  This likely never worked.


botovq:
There doesn't seem to be a supported way of runing a FreeBSD CI on github. Maybe there's something like the Solaris CI that could be done here? Otherwise we'd need to self-host something.

@joshuasing, could you take a look? I twould be great to have some coverage for both autoconf and cmake builds on FreeBSD.

joshuasing (approved):

PR #1109, opened 2024-10-15
ci: add freebsd workflow

joshuasing:
Add a GitHub Actions workflow to run `make check` inside a FreeBSD VM, which runs daily at 00:00 or when manually requested.
This is similar to the Solaris VM workflow, and uses an action from the same `vmactions` org.

Successful CI run:
https://github.com/joshuasing/libressl-portable/actions/runs/11341234538/job/31539126298

Should this also run `gmake test` or similar?

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -1,0 +1,71 @@
+# GitHub Actions workflow to run tests on a FreeBSD VM.
+name: "FreeBSD"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *" # At 00:00 daily.
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.event.number || github.ref }}"
+  cancel-in-progress: true
+
+env:
+  FREEBSD_VERSION: "14.1"
+
+jobs:
+  autoconf:
+    name: "autoconf"
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'libressl' || github.event_name != 'schedule'
+    permissions:
+      contents: read
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v4
+
+      - name: "Setup"
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y automake autoconf libtool
+          ./autogen.sh
+
+      - name: "Build on VM"
+        uses: vmactions/freebsd-vm@v1
+        with:
+          release: "${{ env.FREEBSD_VERSION }}"
+          copyback: false
+          prepare: |
+            pkg install -y autoconf automake libtool
+          run: |
+            ./configure
+            make -j2 check || (cat tests/test-suite.log && exit 1)
+
+  cmake:
+    name: "cmake"
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'libressl' || github.event_name != 'schedule'
+    permissions:
+      contents: read
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v4
+
+      - name: "Setup"
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y automake autoconf libtool
+          ./autogen.sh
+
+      - name: "Build on VM"
+        uses: vmactions/freebsd-vm@v1
+        with:
+          release: "${{ env.FREEBSD_VERSION }}"
+          copyback: false
+          prepare: |
+            pkg install -y cmake ninja
+          run: |
+            export CTEST_OUTPUT_ON_FAILURE=1
+            cmake -G Ninja -B build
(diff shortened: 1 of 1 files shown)


.github/workflows/freebsd.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y automake autoconf libtool
+          ./autogen.sh
+
+      - name: "Build on VM"
+        uses: vmactions/freebsd-vm@v1
+        with:
+          prepare: |
+            pkg install -y gcc gmake
```
botovq, line 34: I would not install gcc. This should be using FreeBSD's clang, which is what `cc` defaults to. There's no need to add `gmake`, but you'll need to `pkg install -y autoconf automake libtool` for the autoconf build and `pkg install cmake ninja` for the cmake build.

botovq (changes requested):
> Should this also run gmake test or similar?

`gmake -j2 check` is already part of the script, so no need for this.

What's missing is the cmake stanza. Something like this

```console
$ cmake -G Ninja -B build
$ ninja -C build
$ ninja -C build test
```

.github/workflows/freebsd.yml
```diff
@@ -31,7 +31,12 @@ jobs:
         uses: vmactions/freebsd-vm@v1
         with:
           prepare: |
-            pkg install -y gcc gmake
+            pkg install -y autoconf automake libtool
+            pkg install -y cmake ninja
           run: |
             MAKE=gmake ./configure
             gmake -j2 check || (cat tests/test-suite.log && exit 1)
```
botovq, line 38: sorry I should've been clearer. `gmake` needs to be replaced with `make` throughout

.github/workflows/freebsd.yml
```diff
@@ -31,7 +31,12 @@ jobs:
...
+            
+            cmake -G Ninja -B build
+            ninja -C build
+            ninja -C build test
```
botovq, line 42: It would be nice to know if only autoconf breaks or only cmake. So could this be split in some kind of matrix?

joshuasing: Hmm, splitting it into a matrix would require running two separate VMs, which is probably fine.

.github/workflows/freebsd.yml
joshuasing: Ah, oops!

.github/workflows/freebsd.yml
botovq: Yes, I think it should be fine. It seems to run quite quickly.

botovq (approved):
Looks good to me now. Do you have a successful run somewhere given that it doesn't run in the per-PR CI?

botovq:
```
The process '/usr/bin/bash' failed with exit code 1
```

Where does that come from? Bash isn't part of FreeBSD. You'd need to `pkg install` it and use `/usr/local/bin/bash` if it's really needed. Better use a posix shell and invoke it as `/bin/sh`.

> Would it be worth setting --output-on-failure or similar to log more information when a test fails?

Yeah, that would make sense. Not sure what the right incantation is, though.

>  I was also going to pin the FreeBSD version and disable "copyback" since we don't need the action to copy files back from the VM, is that okay or should I leave it as-is?

Agreed, I don't think we need that.

botovq:
```
  	 62 - err_test (SEGFAULT)
  	102 - signertest (Failed)
  	117 - tlstest (Failed)
  	122 - verifytest (Failed)
```

The three plain failures are what I saw. Should be fixed by rebasing your branch on top of HEAD. I think I know what the issue with the err_test is, will look.

botovq:
I can't repro the segfault. I'll need to instrument this CI VM to figure it out. Will take a while.

joshuasing:
@botovq I will make the changes I mentioned and will rebase on `master`.

> Where does that come from? Bash isn't part of FreeBSD. You'd need to pkg install it and use /usr/local/bin/bash if it's really needed. Better use a posix shell and invoke it as /bin/sh.
I am not sure. Where are you seeing that? :thinking: 
I didn't see it exit with either of those

botovq:
I think I simply got confused about that bash thing. It's in the annotations "panel" when I click on https://github.com/joshuasing/libressl-portable/actions/runs/11341798906

joshuasing:
Oh, I see. I believe that comes from the action running bash to start the VM :thinking:

botovq:
Agreed.

joshuasing:
Oops, forgot to rebase. I have now rebased, here is a run with the latest commits on `master`: https://github.com/joshuasing/libressl-portable/actions/runs/11343486719

botovq:
Thanks. I'll merge if green.

PR #1119, opened 2024-10-29
cmake: use `CMAKE_INSTALL_SYSCONFDIR`

### LINKED ISSUE #1118: cmake: install `--prefix` is ignored for `/etc/ssl/` files
When installing LibreSSL with `cmake --install`, `--prefix` is ignored for `/etc/ssl/` files:
```shell
cmake -B _bld
cmake --build _bld
cmake --install _bld --prefix _out
```
→
```
[...]
-- Installing: /path/to/libressl/_out/lib/pkgconfig/libtls.pc
-- Installing: /usr/local/etc/ssl/cert.pem
-- Installing: /usr/local/etc/ssl/openssl.cnf
-- Installing: /usr/local/etc/ssl/x509v3.cnf
-- Installing: /usr/local/etc/ssl/certs
```

It overwrites system files or fails, while missing these files from the requested prefix.

It's possible to work around this by using the `DESTDIR` env instead.

vszakats:
Replace `${CMAKE_INSTALL_PREFIX}/etc` with
`${CMAKE_INSTALL_SYSCONFDIR}`.

It makes the install step honor the `--prefix` option.

Fixes #1118


botovq:
Great. Thanks!

PR #1120, opened 2024-11-02
Assume unknown operating systems use ELF.

sortie:
There's a lot of new Unix-like operating systems out that that might port LibreSSL (my Sortix is among them) and they all use ELF. If the operating system doesn't use ELF or isn't an Unix, it doesn't have a chance at working out of this box anyway, and this change makes LibreSSL work on a generic and sufficiently featured POSIX system.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -32,10 +32,6 @@ LT_INIT([pic-only])
 
 CHECK_OS_OPTIONS
 
-if test "$HOST_OS" = "unsupported"; then
-	AC_MSG_ERROR([unsupported platform: $host_os])
-fi
-
 CHECK_C_HARDENING_OPTIONS
 
 DISABLE_AS_EXECUTABLE_STACK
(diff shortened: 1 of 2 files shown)


sortie:
I'm open to adding a warning at the end of configure. In an earlier version, I did add a warning, but drowned out in the middle of the output, but it makes sense to put at at the end of configure.

Personally if I port a package to a new platform, I very much want the breakage other people reported, so I can fix it. Artificial allowlists like this, on the other hand, makes porting to new platforms really difficult if one has to send patches to N projects that would otherwise just work out of the box. It's reasonable to expect people to contribute additional configuration if their platforms require unusual tweaks.

sortie:
I added a warning at the very end of configure, I think that'll work for everyone :)

I'd prefer to not register Sortix as an supported operating system just yet, since your upstream sources don't work out of the box yet. I'm in the process of sending my local patches upstream so my patch is smaller :)

sortie:
My patches got sent to openbsd-tech: https://marc.info/?l=openbsd-tech&m=173056673729921&w=2

botovq:
Thanks. Got them in my inbox. Will take a look soon.

PR #1122, opened 2024-11-02
Print uid_t portably with PRIu64.

sortie:
uid_t on Sortix is 64-bit.

I'm unclear if PRIu64 exists on Visual Studio like you suggested if %llu doesn't work. If not, then we could compat header inttypes.h and add a fallback definition for MSVC that surely has another sequence that does it.

... It might be easier to just drop the uids from the error message. "session file has incorrect owner" is clear enough.


botovq:
Thanks.

The PRI dance was necessary for a long time, but I forgot that we dropped
support for the ancient Windows compiler that needed it last year:

https://github.com/libressl/portable/pull/888

I'll land your %llu version instead and see if I can pull this upstream
at some point. No need for another PR.

Dropping the uid info in parens would make the error message inconsistent
with the permission error in the same function, which I think is something
we would want to avoid.

PR #1126, opened 2024-11-30
Added CMake hint when searching for LibreSSL directory on macOS

csb6:
The Homebrew package manager commonly used on macOS places LibreSSL in a known location, so this patch adds that directory to the search list in the CMake find script.

This makes it so macOS users do not have to directly set LIBRESSL_ROOT_DIR if they have installed it using Homebrew.

### DIFF
--- a/FindLibreSSL.cmake
+++ b/FindLibreSSL.cmake
@@ -81,6 +81,11 @@ if (WIN32)
         "${_programfiles}/LibreSSL"
     )
     unset(_programfiles)
+elseif(APPLE)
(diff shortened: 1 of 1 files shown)


botovq:
Thanks, I'm fine with an addition like this. I suppose people using other package managers on macOS (such as fink, pkgsrc, mac ports, nix, ...)  can add their default location to the list of paths if want to have the same convenience. No need to overcomplicate this now.

vszakats:
It could be useful to also add the new default path for ARM machines: `/opt/homebrew/opt/libressl`.

PR #1127, opened 2024-12-17
Hook ML-KEM related tests

nak3:
This patch adds ML-KEM related tests.

PR #1128, opened 2024-12-17
Some fixes for ML-KEM tests

botovq:
- Run the 1024 iteration test as well as unit tests.
- Remove unused test script with unexpected contents.

Follow-on to #1127


nak3:
Thank you for the follow-up 🙏

PR #1129, opened 2024-12-18
unset CC only for mingw and arm32

nak3:
Since `scripts/test` always unset "CC", the github action for "clang"
in Linux does not use clang unintentionally.

This patch unset CC only for mingw{32,64} and arm32.

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -2,7 +2,10 @@
 set -e
 set -x
 
-unset CC
+if [ "$ARCH" = "mingw32" ] || [ "$ARCH" = "mingw64" ] || [ "$ARCH" = "arm32" ]; then
+	unset CC
+fi
+
 ENABLE_ASM="${ENABLE_ASM:=ON}"
 
 if type apt-get >/dev/null 2>&1; then
--- a/scripts/test
+++ b/scripts/test
@@ -2,6 +2,10 @@
 set -e
 set -x
 
+if [ "$ARCH" = "" ]; then
+	ARCH=`uname -m`
+fi
+
 if [ "$ARCH" = "mingw32" -o "$ARCH" = "mingw64" -o "$ARCH" = "arm32" ]; then
 	unset CC
 fi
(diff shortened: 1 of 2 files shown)


nak3:
> Since scripts/test always unset "CC", the github action for "clang"
in Linux does not use clang unintentionally.

It can be confirmed via recent build log like:

https://github.com/libressl/portable/actions/runs/12370930068/job/34525970295
-> `ubuntu-20.04/native (clang)` -> `view raw log`

```
2024-12-17T10:28:50.3399846Z checking if compiling with clang... no
```

scripts/test
```diff
@@ -2,7 +2,10 @@
 set -e
 set -x
 
-unset CC
+if [ "$ARCH" = "mingw32" ] || [ "$ARCH" = "mingw64" ] || [ "$ARCH" = "arm32" ]; then
```
botovq, line 5: Is there a benefit of preferring this over
```suggestion
if [ "$ARCH" = "mingw32" -o "$ARCH" = "mingw64" -o "$ARCH" = "arm32" ]; then
```

nak3: Thank you! Updated.

scripts/test
```diff
@@ -2,7 +2,7 @@
 set -e
 set -x
 
-if [ "$ARCH" = "mingw32" ] || [ "$ARCH" = "mingw64" ] || [ "$ARCH" = "arm32" ]; then
+if [ "$ARCH" = "mingw32" -o "$ARCH" = "mingw64" -o "$ARCH" = "arm32" ]; then
```
botovq, line 5: It is late here... You probably want to move the 'if [ "$ARcH" = "" ]` test  up to before this line so that we can be sure that ARCH is set.

nak3: Thank you! Yes, I moved it to up.

joshuasing:
Nice catch, thank you!

Issue #1025, opened 2024-03-19
libressl-3.9.0: build fails for ios


chipitsine:
I ran build script from https://github.com/build-xcframeworks/libressl

```
./configure --host=x86_64-apple-darwin --prefix=/Users/runner/work/libressl/libressl/build/libressl-build/simulator_x86_64 CC=/usr/bin/clang CPPFLAGS=-I/Applications/Xcode_15.0.1.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator17.0.sdk/usr/include/ 'CFLAGS= -arch x86_64 -miphoneos-version-min=13.0 -pipe -no-cpp-precomp -isysroot /Applications/Xcode_15.0.1.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator17.0.sdk' 'CPP=/usr/bin/cpp ' LD=/Applications/Xcode_15.0.1.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/usr/bin/ld
...
...
/Applications/Xcode_15.0.1.app/Contents/Developer/usr/bin/make  all-am
  CPPAS    aes/libcrypto_la-aes-macosx-x86_64.lo
  CPPAS    aes/libcrypto_la-bsaes-macosx-x86_64.lo
  CPPAS    aes/libcrypto_la-vpaes-macosx-x86_64.lo
  CPPAS    aes/libcrypto_la-aesni-macosx-x86_64.lo
  CPPAS    aes/libcrypto_la-aesni-sha1-macosx-x86_64.lo
  CPPAS    bn/libcrypto_la-modexp512-macosx-x86_64.lo
  CPPAS    bn/libcrypto_la-mont-macosx-x86_64.lo
  CPPAS    bn/libcrypto_la-mont5-macosx-x86_64.lo
  CPPAS    camellia/libcrypto_la-cmll-macosx-x86_64.lo
  CPPAS    md5/libcrypto_la-md5-macosx-x86_64.lo
  CPPAS    modes/libcrypto_la-ghash-macosx-x86_64.lo
  CPPAS    rc4/libcrypto_la-rc4-macosx-x86_64.lo
  CPPAS    rc4/libcrypto_la-rc4-md5-macosx-x86_64.lo
  CPPAS    sha/libcrypto_la-sha1-macosx-x86_64.lo
[...]

PR #1130, opened 2024-12-19
Add in missing endianess functions for Apple targets

palmerc:
Apple SDKs indeed have the endian.h headers, but they do not contain these functions and therefore will error out when doing a cmake build. Remove the conditional HAVE_ENDIAN check or remove the conditional check for Apple compatibility? I chose to remove the conditional check for Apple compatibility.

### DIFF
--- a/include/compat/endian.h
+++ b/include/compat/endian.h
@@ -62,7 +62,7 @@
 #endif
 #endif
 
-#if defined(__APPLE__) && !defined(HAVE_ENDIAN_H)
+#if defined(__APPLE__)
 #include <libkern/OSByteOrder.h>
 #define be16toh(x) OSSwapBigToHostInt16((x))
 #define htobe16(x) OSSwapHostToBigInt16((x))
@@ -74,7 +74,7 @@
 #define htobe64(x) OSSwapHostToBigInt64(x)
 #define le64toh(x) OSSwapLittleToHostInt64(x)
 #define be64toh(x) OSSwapBigToHostInt64(x)
-#endif /* __APPLE__ && !HAVE_ENDIAN_H */
+#endif /* __APPLE__ */
 
 #if defined(_WIN32) && !defined(HAVE_ENDIAN_H)
 #include <winsock2.h>


botovq:
Thanks. Something similar came up here #1025 but this sort of got stuck since nobody could tell me what the correct approach was. Can you tell me if this is good enough to build on iOS as well, i.e., does this address the issue in #1025?

palmerc:
I am building macOS, iPhoneOS and iPhoneSimulator in a single shot

palmerc:
A better approach might be to check for the individual functions, but it is sort of an all-or-nothing affair

botovq:
This endian.h compat thing is just such an unbelievable mess :)

We copied this from openiked, which has always had this `HAVE_ENDIAN_H` define.

Is it aversion of [this Endian.h][1] which is picked up due to case-insensitivity of the file system on macOS (I assume that's still the case)?

All that said, I am inclined to merge this. I will do so in a couple of days.


[1]: https://github.com/mstg/iOS-full-sdk/blob/master/iPhoneOS9.3.sdk/usr/include/Endian.h

palmerc:
All of the current SDKs have these endian files:

```
% find . -name Endian.h    
./usr/include/Endian.h

find . -name endian.h
./usr/include/c++/v1/__bit/endian.h
./usr/include/machine/endian.h
./usr/include/arm/endian.h
```

palmerc:
I had failed to update the comment at the bottom of the block to match, but I've done that now.

botovq:
Thanks. I was hoping we could avoid doing in this super fine grained manner, but yes, it's probably cleaner than what we have now... It's just surprising that I can't find a ready-made solution that covers all our needs. It feels like it should exist...

palmerc:
What if I just switched the check to this:
```
if(NOT APPLE AND NOT WIN32)
	check_include_files(endian.h HAVE_ENDIAN_H)
	if(HAVE_ENDIAN_H)
		add_definitions(-DHAVE_ENDIAN_H)
	endif()
endif()
```

botovq:
I'm pretty sure that the previous diff was better. This logic introduces an unexpected difference between the CMake and automake builds which will likely bite us down the road.

I'd like to merge the previous version and next time an issue shows up we'll have to sit down and do the grunt work of solving this "the right way."

palmerc:
Doesn't the other formulation leave windows builds broken? In endian.h, right after the equivalent Apple bit there is similar code defining non-existent functions with the expectation that endian.h does not exist. I can't answer whether or not windows machines have endian.h or not, nor whether they are missing these #defines.

```
#if defined(_WIN32) && !defined(HAVE_ENDIAN_H)
#include <winsock2.h>

#define be16toh(x) ntohs((x))
#define htobe16(x) htons((x))
#define le32toh(x) (x)
#define be32toh(x) ntohl((x))
#define htole32(x) (x)
#define htobe32(x) ntohl((x))
#define be64toh(x) ntohll((x))

#if !defined(ntohll)
#define ntohll(x)                                                              \
  ((1 == htonl(1))                                                             \
       ? (x)                                                                   \
       : ((uint64_t)ntohl((x)&0xFFFFFFFF) << 32) | ntohl((x) >> 32))
#endif
#if !defined(htonll)
#define htonll(x)                                                              \
  ((1 == ntohl(1))                                                             \
       ? (x)                                                                   \
       : ((uint64_t)htonl((x)&0xFFFFFFFF) << 32) | htonl((x) >> 32))
#endif

#define htobe64(x) ntohll((x))
#define htole64(x) (x)
#define le64toh(x) (x)
#endif /* _WIN32 && !HAVE_ENDIAN_H */
```

palmerc:
In any case, the original version works for Apple and that is what matters to me :)

botovq:
No, it's all good. Merged now, thanks.

PR #1134, opened 2025-01-09
What happens on ubuntu-24.04?

botovq:
Actions have been issuing annoying warnings about ubuntu-24.04 becoming the default for ubuntu-latest for some definition of "soon". Let's see if we care.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -10,7 +10,7 @@ concurrency:
 jobs:
   test:
     name: "Test ${{ matrix.name }}"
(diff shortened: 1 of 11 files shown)


joshuasing (approved):
GitHub is currently rolling out 24.04 for `ubuntu-latest`, so the warnings seem a bit late haha.

Looks good to me, and it's probably best we pin these anyway.

botovq:
@joshuasing that's what I thought, thanks. Should we pin the other -latest as well? if so, could you send a PR, please?

PR #1135, opened 2025-01-11
ci: improve cmake config workflow

joshuasing:
Tidy up the CMake config check workflow:
1. Follow same style as other workflows
2. Set `strategy.fail-fast: true`, so that all workflow runs for a ref are cancelled if one OS fails (is this correct?)
3. Restrict token permissions to `contents: read`.
4. Pin runner images to versions, instead of using `windows-latest` and `macos-latest`.
5. Rename file `cmake_config.yml` -> `cmake-config.yml` to match other workflows.


botovq (approved):
For the future, it would be nice if you could make the rename a separate commit. This would make sure the actual changes are visible in diff format rather than having to resort to doing this by hand.

Likewise, I wouldn't mind having cosmetics like adding quotes and pure whitespace changes in separate commits. This makes it much easier to focus on the actual changes.

joshuasing:
> For the future, it would be nice if you could make the rename a separate commit. This would make sure the actual changes are visible in diff format rather than having to resort to doing this by hand.
> 
> Likewise, I wouldn't mind having cosmetics like adding quotes and pure whitespace changes in separate commits. This makes it much easier to focus on the actual changes.

Sure, will do. Sorry for the inconvenience.
Are separate commits in one pull request okay, or would separate pull requests (with one commit per pull request) be better?

botovq:
A single PR is ok, containing a stack of commits where each commit does a single change each.

PR #1145, opened 2025-04-07
Fix no pointer sign warning on C++ compilers

mustafa-gokce:


CMakeLists.txt
```diff
@@ -131,7 +131,7 @@ add_definitions(-D__END_HIDDEN_DECLS=)
 set(CMAKE_POSITION_INDEPENDENT_CODE true)
 
 if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
-	add_definitions(-Wno-pointer-sign)
+	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-pointer-sign")
```
botovq, line 134: While this makes sense, I wonder if this should not be using `add_compile_options(-Wno-pointer-sign)` instead.

Also (probably to be considered separately), what about the `add_definitions(-fno-common)` a bit further up in the file?

Issue #1123, opened 2024-11-05
Support Loongarch64 (error: `fatal error: crypto_arch.h: No such file or directory`)?


abcfy2:
Hi there. Do you plan to support `loongarch64` ? Since openssl already support this arch.

Toolchain: https://github.com/musl-cross/musl-cross/releases/download/20241103/loongarch64-unknown-linux-musl.tar.xz

configure options:

```sh
./configure --host=loongarch64-unknown-linux-musl --prefix=/cross_root/loongarch64-unknown-linux-musl --enable-silent-rules --enable-static --disable-shared --with-openssldir=/etc/ssl
```

Error:

```
Making all in include
make[1]: Entering directory '/usr/src/libressl-4.0.0/include'
Making all in openssl
make[2]: Entering directory '/usr/src/libressl-4.0.0/include/openssl'
echo "generating opensslconf.h ..."
generating opensslconf.h ...
make  all-am
make[3]: Entering directory '/usr/src/libressl-4.0.0/include/openssl'
echo "generating opensslconf.h ..."
generating opensslconf.h ...
make[3]: Leaving directory '/usr/src/libressl-4.0.0/include/openssl'
make[2]: Leaving directory '/usr/src/libressl-4.0.0/include/openssl'
make[2]: Entering directory '/usr/src/libressl-4.0.0/include'
make[2]: Nothing to be done for 'all-am'.
make[2]: Leaving directory '/usr/src/libressl-4.0.0/include'
make[1]: Leaving directory '/usr/src/libressl-4.0.0/include'
Making all in crypto
make[1]: Entering directory '/usr/src/libressl-4.0.0/crypto'
echo "generating crypto_portable.sym ..."
generating crypto_portable.sym ...
cp ../crypto/crypto.sym crypto_portable.sym
chmod u+w crypto_portable.sym
make  all-am
make[2]: Entering directory '/usr/src/libressl-4.0.0/crypto'
  CC       libcrypto_la-cpt_err.lo
  CC       libcrypto_la-cryptlib.lo
  CC       libcrypto_la-crypto_ex_data.lo
  CC       libcrypto_la-crypto_init.lo
  CC       libcrypto_la-crypto_lock.lo
  CC       libcrypto_la-cversion.lo
  CC       libcrypto_la-malloc-wrapper.lo
  CC       libcrypto_la-mem_clr.lo
  CC       libcrypto_la-mem_dbg.lo
  CC       libcrypto_la-o_fips.lo
[...]

PR #1146, opened 2025-04-10
Basic loongarch64 support

iv-m:
Just wire things up to make the project buildable on LoongArch64.

opensslconf.h is taken form riscv64 folder as is, and is the same as on some other 64 bit architectures.

References: https://github.com/libressl/portable/issues/1123


iv-m:
This builds on my loongarch64 machine (Loongson-3A6000-HV, XA61200 board), both with autotools

```
$ ./autogen.sh
$ ./configure
$ ./make -j8 check
```

and with `cmake`

```
$ ./autogen.sh
$ mkdir build && cd build
$ cmake ..
$ make -j8
$ make test
```

All tests are passing.

Let me know if I there is anything else I should do here, e.g. updating changelog maybe?

joshuasing:
> @joshuasing could you see if you find a way to test this? Perhaps there's a way to run it in a vm like we do for freebsd?

Sure! I'll have a look into it, it seems possible but I don't know of a way off the top of my head.

tests/CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ elseif(HOST_I386)
 	include_directories(../crypto/arch/i386/)
 elseif(HOST_MIPS64)
 	include_directories(../crypto/arch/mips64)
+elseif(HOST_LOONGARCH64)
+	include_directories(../crypto/arch/loongarch64)
```
botovq, line 34: Please move this up above MIPS64

botovq (approved):
This looks good to me. I hope Joshua will figure out a way to give us some minimal test coverage, but I'm fine with landing it. It's not going to get in the way.

busterb:
Looks fine to me too.

I think we could possibly use qemu's support for binaries cross-compiled with binfmt support loaded to test. That's what I do for testing cross architectures locally. Not sure if github actions will allow it, but I seem to remember it working in the past.

busterb (approved):

busterb:
It's a little more difficult in that disabling static builds doesn't quite work:
```
$ ./configure --host=loongarch64-linux-gnu --enable-static --disable-shared
checking build system type... ^C
$ file ./apps/openssl/openssl 
./apps/openssl/openssl: ELF 64-bit LSB pie executable, LoongArch, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-loongarch-lp64d.so.1, BuildID[sha1]=0e11eaf674e9e3b4049f14b87c652341946c1b94, for GNU/Linux 5.19.0, with debug_info, not stripped
```

but Debian doesn't quite support it as a mainline architecture, even though the qemu emulator and compilers are there:
```
$ ./apps/openssl/openssl 
loongarch64-binfmt-P: Could not open '/lib64/ld-linux-loongarch-lp64d.so.1': No such file or directory
```

`dpkg --add-architecture loong64` would be nice if it worked out of the box so qemu could find the loader. Maybe there's a way with some env variables and/or symlinks today.

tests/CMakeLists.txt
botovq: Fixed this in a follow-on. Thanks for the PR!

PR #1148, opened 2025-04-12
ci: test on loongarch64

joshuasing:
Add support for testing on `loongarch64` in the `/scripts/test` script and in GitHub Actions. This works by using GCC cross-compilation and QEMU to build and run the tests.

Currently compilation on `master` will fail, so this should be merged only after #1146 is merged.

Tested locally on #1146, all tests pass and the `file` output shows LoongArch:
```
============================================================================
Testsuite summary for libressl 4.1.0
============================================================================
# TOTAL: 136
# PASS:  136
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
make[3]: Leaving directory '/libressl/tests'
make[2]: Leaving directory '/libressl/tests'
make[1]: Leaving directory '/libressl/tests'
make[1]: Entering directory '/libressl'
make[1]: Nothing to be done for 'check-am'.
make[1]: Leaving directory '/libressl'
+ file apps/openssl/openssl
apps/openssl/openssl: ELF 64-bit LSB executable, LoongArch, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-loongarch-lp64d.so.1, BuildID[sha1]=f666f5363db4c2f555f71d916ee459221f810e42, for GNU/Linux 5.19.0, with debug_info, not stripped
```

Note: The script uses "loong64" instead of "loongarch64". "loong64" is used by Go (where support is maintained by Loongson), Debian, etc. and is shorter. I don't mind changing it if anyone thinks we should use the longer name.

Also, I am not too familiar with automake/autoconfig/gcc, so if there's anyway this can be improved, please let me know.

Related to: #1146, #1123

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -134,7 +134,7 @@ elif [ "$ARCH" = "arm32" -o "$ARCH" = "arm64" ]; then
 
 elif [ "$ARCH" = "loong64" ]; then
 	sudo apt install -y qemu-user-static binfmt-support g++-14-loongarch64-linux-gnu
-	sudo ln -sf /usr/loongarch64-linux-gnu/lib/ld-linux-loongarch-lp64d.so.1 /lib
+	sudo ln -sf /usr/loongarch64-linux-gnu/lib64/ld-linux-loongarch-lp64d.so.1 /lib64
 
 	CC=loongarch64-linux-gnu-gcc-14
 	CXX=loongarch64-linux-gnu++-14
--- a/scripts/test
+++ b/scripts/test
@@ -132,6 +132,21 @@ elif [ "$ARCH" = "arm32" -o "$ARCH" = "arm64" ]; then
 
 	file apps/openssl/.libs/openssl
 
+elif [ "$ARCH" = "loong64" ]; then
+	sudo apt install -y qemu-user-static binfmt-support g++-14-loongarch64-linux-gnu
+	sudo ln -sf /usr/loongarch64-linux-gnu/lib/ld-linux-loongarch-lp64d.so.1 /lib
+
+	CC=loongarch64-linux-gnu-gcc-14
+	CXX=loongarch64-linux-gnu++-14
+	AR=loongarch64-linux-gnu-ar
+	STRIP=loongarch64-linux-gnu-strip-14
+	RANLIB=loongarch64-linux-gnu-ranlib
+
+	./configure --host=loongarch64-linux-gnu
+	LD_LIBRARY_PATH=/usr/loongarch64-linux-gnu/lib make -j 4 check
+
+	file apps/openssl/openssl
+
 elif [ "$ARCH" = "mips32" -o "$ARCH" = "mips64" ]; then
 	sudo apt-get install -y qemu-user-static binfmt-support
 
--- a/scripts/test
+++ b/scripts/test
@@ -133,13 +133,13 @@ elif [ "$ARCH" = "arm32" -o "$ARCH" = "arm64" ]; then
 	file apps/openssl/.libs/openssl
 
 elif [ "$ARCH" = "loong64" ]; then
-	sudo apt install -y qemu-user-static binfmt-support g++-loongarch64-linux-gnu
+	sudo apt install -y qemu-user-static binfmt-support g++-14-loongarch64-linux-gnu
 	sudo ln -sf /usr/loongarch64-linux-gnu/lib/ld-linux-loongarch-lp64d.so.1 /lib
 
-	CC=loongarch64-linux-gnu-gcc
-	CXX=loongarch64-linux-gnu++
+	CC=loongarch64-linux-gnu-gcc-14
+	CXX=loongarch64-linux-gnu++-14
 	AR=loongarch64-linux-gnu-ar
-	STRIP=loongarch64-linux-gnu-strip
+	STRIP=loongarch64-linux-gnu-strip-14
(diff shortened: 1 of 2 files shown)


botovq:
Ah that's nice, thanks. I'll merge #1146 in a few minutes and will ping you to rebase on top of that.

busterb:
This suggests we should add powerpc64 and riscv64 tests too, if we want to cover the other Linux architectures that are missing CI tests.

botovq:
ok you can rebase now.

> This suggests we should add powerpc64 and riscv64 tests too, if we want to cover the other Linux architectures that are missing CI tests.

Makes sense to me. Slightly less urgent since these are run tested in openbsd.

botovq:
I'll wait for this to land and will then rebase and land my mips32 PR #1147 since this will also conflict.

joshuasing:
Okay, rebased. If we can use the same setup for riscv64/powerpc64 that would be great. I can poke around to try and get those working too. I also want to do a bit of tidy up in the test script, but that can be separate.

scripts/test
```diff
@@ -132,6 +132,21 @@ elif [ "$ARCH" = "arm32" -o "$ARCH" = "arm64" ]; then
 
 	file apps/openssl/.libs/openssl
 
+elif [ "$ARCH" = "loong64" ]; then
+	sudo apt install -y qemu-user-static binfmt-support g++-loongarch64-linux-gnu
+	sudo ln -sf /usr/loongarch64-linux-gnu/lib/ld-linux-loongarch-lp64d.so.1 /lib
+
+	CC=loongarch64-linux-gnu-gcc
```
busterb, line 139: These variable assignments are not really doing anything; should be able to remove. --host on configure will figure this all out automatically. Look at how mips and arm doit.

joshuasing: I was unable to get it to build locally without these, it always built for x86-64 instead. It may be an issue just on my machine though, I can try removing to see if it works properly in GitHub Actions.

scripts/test
```diff
@@ -132,6 +132,21 @@ elif [ "$ARCH" = "arm32" -o "$ARCH" = "arm64" ]; then
 
 	file apps/openssl/.libs/openssl
 
+elif [ "$ARCH" = "loong64" ]; then
+	sudo apt install -y qemu-user-static binfmt-support g++-14-loongarch64-linux-gnu
```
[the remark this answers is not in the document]
joshuasing: It ran into the same issue I had locally, `g++-loongarch64-linux-gnu` doesn't appear to exist in the Ubuntu universe repo, whereas `g++-14-loongarch64-linux-gnu` does.

scripts/test
botovq: If dropping it works I'm fine with that, but I have definitely seen cross-compilation fail due to using the host's default compiler (it's been a while - macos arm64 seems to have hit this from time to time).

botovq:
I'd like to see something land right now that works. We can make it pretty later and put it on our todo lists for next month.

joshuasing:
https://github.com/libressl/portable/pull/1148#discussion_r2040654333

I pushed a commit to change from `g++-14-loongarch64-linux-gnu` to `g++-loongarch64-linux-gnu`, however it seems `g++-loongarch64-linux-gnu` does not exist in the Ubuntu universe repositories. It also fails using the ubuntu Docker image locally (with the universe repo added).

`g++-14-loongarch-linux-gnu` exists on both Ubuntu 22.04 and 24.04, so I think this works for now, but it would be ideal to use the non-versioned package :thinking:

joshuasing:
Never mind, it looks like `universe` repository is also missing :thinking:

joshuasing:
Okay, it is working in CI now. I ran into a few issues:
1. Ubuntu 22.04 does not have a gcc/g++ package for `loongarch64` cross-compilation, only Ubuntu 24.04 and newer does.

2. Ubuntu 24.04 doens't have a `gcc++-loongarch64-linux-gnu` package, only `gcc++-{14,13}-loongarch64-linux-gnu` packages, it seems. Ubuntu 24.10 seems to have the non-versioned package.

3. For some reason, locally running the test script in an Ubuntu 24.04 Docker container, `/lib/ld-linux-loongarch-lp64d.so.1` is used - however, in GitHub Actions, `/lib64/ld-linux-loongarch-lp64d.so.1` is used. I have no clue why this differs. I have changed the script to use `/lib64`, however this breaks the script locally... Maybe we can setup both?

botovq:
> Maybe we can setup both?

I see nothing wrong with adding two symlinks. If that's what you mean. Also, `LD_LIBRARY_PATH` works like PATH with colon separation if that should need tweaks, too.

botovq (approved):
I'm happy to merge this as it is, but I can also wait if you want to try fixing the lib vs lib64 issue.

joshuasing:
> I'm happy to merge this as it is, but I can also wait if you want to try fixing the lib vs lib64 issue.

Sorry for the delay. The latest commit seems to work consistently in GitHub Actions as well as on my machine.

When running in Docker (I was trying to better replicate an actions runner and figure out which packages would be needed), `ARCH=loong64 ./scripts/test` always builds for x86_64 using `/lib64/ld-linux-x86-64.so.2`, seemingly on any commit with either lib/lib64. It had worked previously, so I have no clue why it suddenly isn't now...

PR #1150, opened 2025-04-12
Alt fix for #1069

busterb:
Just doing some testing to see if we can implement a fix for #1069 on top of #1071 using method suggested in https://github.com/libressl/portable/issues/266#issuecomment-2227664358

There are a handful of tests that already are failing in master on wine + mingw-w64, but pushing this to check other platforms through CI first.

### DIFF
--- a/include/compat/sys/stat.h
+++ b/include/compat/sys/stat.h
@@ -118,4 +118,12 @@
 
 #endif
 
+#ifdef _WIN32
+int posix_fstat(int fd, struct stat *statbuf);
+
+#ifndef NO_REDEF_POSIX_FUNCTIONS
+#define fstat(fd, statbuf) posix_fstat(fd, statbuf)
+#endif
+#endif
+
 #endif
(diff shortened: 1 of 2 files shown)


busterb:
No new issues found in local testing. Unless there are any objections, going to try land this tomorrow. Thanks for the suggestions @Ono-Sendai and @datadiode

include/compat/sys/stat.h
```diff
@@ -118,4 +118,12 @@
 
 #endif
 
+#ifdef _WIN32
+int posix_fstat(int fd, struct stat *statbuf);
+
+#ifndef NO_REDEF_POSIX_FUNCTIONS
+#define fstat(fd, statbuf) posix_fstat(fd, statbuf)
```
botovq, line 125: This is fine for now, but I think we should consider switching things to a `libressl_*` prefix also for windows. Similarly as here:

https://github.com/libressl/portable/blob/73779a46bf49c4f53cc4b81993135a7408a01963/include/compat/stdio.h#L21-L39

botovq (approved):
Looks good to me, thanks.

datadiode:
@busterb Did you recheck the case from https://github.com/libressl/portable/pull/1071#issuecomment-2409624670?

PR #1164, opened 2025-05-18
experiment with removing BIO_s_log in advance of upstream removal

busterb:
just a test if we disabled it everywhere in advance of it being removed upstream


botovq (approved):
I think you can just land this and I'll clean up the rest when I bump the major.

PR #1166, opened 2025-05-18
Gitignore fixes

busterb:
/crypto/* masks too much and breaks the exclusion function.

The docs for gitignore state that if a parent directory is ignored, negating a pattern with a ! prefix has no effect since it is not possible to reinclude a file if the parent directory is excluded.

https://git-scm.com/docs/gitignore

This just is more explicit, which catches files added out of tree more readily.

Also delete the long-removed libtls-standalone references


botovq:
On Sun, May 18, 2025 at 04:48:08AM -0700, Brent Cook wrote:
> /crypto/* masks too much and breaks the exclusion function.
> 
> The docs for gitignore state that if a parent directory is ignored, negating a pattern with a ! prefix has no effect since it is not possible to reinclude a file if the parent director is excluded.
> 
> https://git-scm.com/docs/gitignore

I no longer use git, so whatever works for you, I can deal with it.

PR #1170, opened 2025-05-22
update.sh: stop generating vpaes and bsaes x86_64 assembly

nak3:
Remove generation of vpaes and bsaes assembly sources from update.sh, as these are removed from source tree.

See: https://github.com/openbsd/src/commit/824f363240372afd9c86bea543128d656b531c16


botovq (approved):
@busterb, I'll let you merge

PR #1174, opened 2025-05-31
Update Windows CI runner to 2025

### LINKED ISSUE #1172: Update Windows 2019 runner
Just a reminder so I don't forget until I unbreak the portable repo due to recent churn.

botovq:
Fixes #1172

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -21,14 +21,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["windows-2022", "windows-2019"]
+        os: ["windows-2022", "windows-2025"]
         arch: ["ARM64", "x64", "Win32"]
         shared: ["ON", "OFF"]
         include:
           - os: "windows-2022"
             generator: "Visual Studio 17 2022"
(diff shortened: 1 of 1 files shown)


busterb (approved):

busterb:
One minor thing to note I've been working on the side on - i386 / arm64 on Windows haven't been actually building the right target, and has really just been doing amd64 regardless due to [CMAKE_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html) returning the wrong value on this platform. If this update breaks or fixes that, we could probably remove those targets for now and reenable later.

busterb:
Also, I'm not sure cmake is new enough here; so might just disable 2025 for now.

botovq:
Yeah, it didn't like 'Visual Studio 17 2025'. I'm trying with the old value we used for 2022 for now, otherwise I'll drop it.

botovq:
> - i386 / arm64 on Windows haven't been actually building the right target

I'm not sure how I can tell. Let's keep this cleanup for a separate PR.

PR #1176, opened 2025-06-02
Automake: use relative paths when linking object lists

cierny:
This PR updates the automake-exported object lists to use relative paths instead of absolute paths.

This solves an issue when building the project from a directory with a longer absolute path (such as within a Bazel project in my use case), which can lead to an `Argument list too long` error due to the large amount of objects with long absolute paths given to `libtool`.


botovq (approved):
Thanks. I'm ok with this, but I'd like to hear from @busterb if he sees a reason not to change that.

busterb:
6 in one half, half dozen in the other for most use cases. This might fix #1024 as well

busterb (approved):

PR #1177, opened 2025-06-04
adjust CMAKE_SYSTEM_PROCESSOR on Windows builds

busterb:
Match CMAKE_GENERATOR_PLATFORM so that cross-builds work properly on Windows Visual Studio targets. Otherwise, we're just targeting the architecture of the build system itself, since CMAKE_SYSTEM_PROCESSOR is not guaranteed to match the target architecture on all platforms. Why CMake doesn't take care of this itself is a mystery! The docs just say 'maybe' and leave the rest as an exercise for the reader: https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html

I'm getting some weird behavior locally with this fix though running in a VM with a shared filesystem, pushing to see if it's something with my setup.


busterb:
Note; it seems like the code to copy the right opensslconf.h system architecture file doesn't always execute in time, especially in multi-threaded builds. This is something I'd also like to fix (maybe it can be executed at configure time) since it makes my local builds a little unreliable.

.github/workflows/windows.yml
```diff
@@ -26,9 +26,8 @@ jobs:
         shared: ["ON", "OFF"]
         include:
           - os: "windows-2022"
-            generator: "Visual Studio 17 2022"
+            generator: "Visual Studio 16 2019"
```
botovq, line 29: We used to check VS 17 2022 on Windows 2022 before I disabled to 2019 VMs, so this kind of goes backwards. The 2025 entry below is a guess which happened to work in the CI.

busterb: Yeah, I wanted to check if the older compiler was at least still available on that platform since that's still supported by MS. I have VS 16 and VD 17 installed locally on my test VMs. They're independent of the OS you run them on. There's no newer VS version than VS 17, so it's kind of redundant otherwise to run this on 2 different OSes. We can probably just choose whether to CI build them all on windows-2022 or 2025.

busterb: Yeah, looks like they just don't install the older toolchain in parallel on the newer Github CI images. I'll back this out.

busterb:
> > This is something I'd also like to fix (maybe it can be executed at configure time) since it makes my local builds a little unreliable.
> 
> Ideally we'd get rid of all the MD garbage in these headers. It's all about legacy stuff like IDEA/RC2/RC4/DES that could probably well do without optimization shenanigans.

My local problem turned out to be a case-sensitive filesystem. A red herring for this, but agree on the longterm goal!

botovq:
> My local problem turned out to be a case-sensitive filesystem being incompatible with CMake's VS generator

Ugh that's nasty.

> but agree on the longterm goal!

working on it :)

PR #1179, opened 2025-06-17
Clean up handling of ASM support code

botovq:
This avoids the patches I had to add to the `aes_amd64.c` and `aes_i386.c` files and stops mixing the `.c` and `.S` sources. I think it's ovreall a bit cleaner than before. Let me know if I missed anything. I'm especially unsure about the `EXTRA_DIST` stuff...


botovq:
@busterb any opinion this? Otherwise I'll merge in a couple of days.

busterb:
You can tell if 'EXTRA_DIST' is needed by running ./dist.sh. Generally as long a file is referenced otherwise in the build config, it's not needed. Let me give it a test.

busterb (approved):
Looks fine, remove the EXTRA_DIST and should be good to go.

PR #1181, opened 2025-07-06
Limit usage of warning flags to C and C++

d3x0r:
`add_compile_options()` adds options to all compilers CMake can invoke, including ASM.  The format of this warning flag causes MASM to fail if the option is specified.  This limits specifying the warning suppression to just C and C++.


botovq:
Thanks. Is there a reason we need this for C++? We don't have any C++ code.

botovq (approved):

Issue #1189, opened 2025-07-30
CMake build doesn't properly define HAVE_RESOLV_H


Un1q32:
the autoconf script checks for resolv.h and defines HAVE_RESOLV_H if it is present, the cmake build doesn't and as a result include/compat/resolv.h does not include the OS resolv.h.

PR #1190, opened 2025-08-01
check for resolv.h and arpa/nameser.h

Un1q32:
The CMake build does not define these macros, autoconf build does.  This fixes the CMake build and brings it on par with autoconf.

Fixes https://github.com/libressl/portable/issues/1189

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -378,6 +378,16 @@ if(HAVE_NETINET_IP_H)
 	add_definitions(-DHAVE_NETINET_IP_H)
 endif()
 
+check_include_files(resolv.h HAVE_RESOLV_H)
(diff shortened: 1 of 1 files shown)


botovq:
Thanks. Looks good to me. I'll merge once the CI is green.

botovq (approved):

Un1q32:
grepping the include/compat directory for "HAVE_*_H" shows there is also a HAVE_READPASSPHRASE_H, but that macro isn't even defined by the autoconf build.  It doesn't cause me any build issues so I'm not gonna touch it.

botovq:
Thanks for pointing it out. That's something that looks pretty harmless but should be fixed. I opened #1191 for that.

PR #1198, opened 2025-08-28
Fix inconsistencies in accept4.c

imorgenstern:
Fix inconsistencies in accept4.c. If the underlying accept() fails the shim returns the listening socket s instead of −1.


busterb:
These look correct, thanks @imorgenstern

PR #1206, opened 2025-10-15
cmake: use CMAKE_INSTALL_*DIR variables as are, if absolute

doronbehar:
These variables can be set to absolute paths, in that case ${prefix} shouldn't be prepended. CMAKE_INSTALL_FULL_*DIR variables could have also been used, but that would hard-code the relativity to the prefix pkg-config variable. See also:

https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html


CMakeLists.txt
```diff
@@ -568,9 +568,9 @@ if(ENABLE_LIBRESSL_INSTALL)
 	if(NOT MSVC)
 		# Create pkgconfig files.
 		set(prefix      ${CMAKE_INSTALL_PREFIX})
-		set(exec_prefix \${prefix})
-		set(libdir      \${exec_prefix}/${CMAKE_INSTALL_LIBDIR})
-		set(includedir  \${prefix}/${CMAKE_INSTALL_INCLUDEDIR})
+		set(exec_prefix ${CMAKE_INSTALL_FULL_BINDIR})
+		set(libdir      ${CMAKE_INSTALL_FULL_LIBDIR})
+		set(includedir  ${CMAKE_INSTALL_FULL_INCLUDEDIR})
```
vszakats: I wonder what is benefit of this change?

The existing solution generates `.pc` files that allow relocation, while
this proposal hard-codes them to absolute paths, losing the dynamic
relationship between values, and preventing relocation. It also makes
the `prefix` item unused, though it still sets it.

If the goal is passing down absolute paths if those were manually
customized as such, this alternative would keep existing functionality, 
while also honoring absolute paths, when necessary:

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -572,8 +572,16 @@ if(ENABLE_LIBRESSL_INSTALL)
 		# Create pkgconfig files.
 		set(prefix      ${CMAKE_INSTALL_PREFIX})
 		set(exec_prefix \${prefix})
-		set(libdir      \${exec_prefix}/${CMAKE_INSTALL_LIBDIR})
-		set(includedir  \${prefix}/${CMAKE_INSTALL_INCLUDEDIR})
+		if(IS_ABSOLUTE ${CMAKE_INSTALL_LIBDIR})
+			set(libdir      ${CMAKE_INSTALL_LIBDIR})
+		else()
+			set(libdir      \${exec_prefix}/${CMAKE_INSTALL_LIBDIR})
+		endif()
+		if(IS_ABSOLUTE ${CMAKE_INSTALL_INCLUDEDIR})
+			set(includedir  ${CMAKE_INSTALL_INCLUDEDIR})
+		else()
+			set(includedir  \${prefix}/${CMAKE_INSTALL_INCLUDEDIR})
+		endif()
 		if(PLATFORM_LIBS)
 			string(REGEX REPLACE ";" " -l" PLATFORM_LDADD ";${PLATFORM_LIBS}")
 		endif()
```

PR #1208, opened 2025-10-21
Rudimentary IRIX support added

Sunny-Maxis:
Hello Libressl team,

This is an attempt by my employer, Kaz Kuroi, to add IRIX support upstream to libressl. This doesn't solve all IRIX related breakage, but it greatly eases porting for others. 

Let me explain my modifications:

arc4random_irix.h - IRIX doesn't support anonymous maps. We get around this using an fd to /dev/zero. Works perfectly fine. 
mman.h - Issue a warning instead of a #error. IRIX is closed source. It won't support anonymous mapping, likely ever. A warning to the builder is sufficient, IMHO. 

getentropy_irix.c - IRIX uses yarrow for /dev/random and urandom, and the fallback hack is just there if there's not sufficient bits in urandom. 

I did not try to patch automake. Please also note, I'm submitting this code on behalf of Kaz, who doesn't use GitHub for personal reasons. If there's questions, it may take me some time to respond. 

Thanks!

Patrick Maxis

### DIFF
--- a/crypto/compat/getentropy_irix.c
+++ b/crypto/compat/getentropy_irix.c
@@ -121,9 +121,9 @@ getentropy(void *buf, size_t len)
         remaining -= take;
     }
     /* cleanup */
-    memset(pool, 0, sizeof(pool));
-    memset(digest, 0, sizeof(digest));
-    memset(&ctx, 0, sizeof(ctx));
+    explicit_bzero(pool, sizeof(pool));
+    explicit_bzero(digest, sizeof(digest));
+    explicit_bzero(&ctx, sizeof(ctx));
 
     errno = save_errno;
     return (0);


crypto/compat/arc4random_irix.h
```diff
@@ -0,0 +1,66 @@
+
```
botovq: We need a license.

crypto/compat/arc4random_irix.h
```diff
@@ -0,0 +1,66 @@
...
+_rs_allocate(struct _rs **rsp, struct _rsx **rsxp)
+{
+        int fd;
+        fd = open("/dev/zero", O_RDWR);
+        if ((*rsp = mmap(NULL, sizeof(**rsp), PROT_READ|PROT_WRITE,
+            MAP_PRIVATE, fd, 0)) == MAP_FAILED)
+                return (-1);
+
+        if ((*rsxp = mmap(NULL, sizeof(**rsxp), PROT_READ|PROT_WRITE,
+            MAP_PRIVATE, fd, 0)) == MAP_FAILED) {
```
botovq: again, don't leak fd

botovq (changes requested):
Thanks. Apologies for never getting back to reviewing the diffs.

As I believe I mentioned on the list when these diffs were first posted, we are not in principle opposed to adding patches to make downstreams' lives easier, but we cannot maintain this ourselves. 

Is there a way to check this in CI to reduce the risk of accidental breakage?

I will have to consult with others if the lack of `MAP_ANON` support in IRIX should be considered a blocker for merging this. 

Some initial feedback below.

include/compat/sys/mman.h
```diff
@@ -11,8 +11,8 @@
 #ifndef MAP_ANON
 #ifdef MAP_ANONYMOUS
 #define MAP_ANON MAP_ANONYMOUS
-#else
-#error "System does not support mapping anonymous pages?"
+#else 
+#warning "System does not support mapping anonymous pages. Build may fail" /* Warn, don't error out. arc4random_irix doesn't use anonymous pages */
```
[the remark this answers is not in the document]
Sunny-Maxis: Our compiler will halt if #error preprocessing directive is supplied. I can add an #else if for __sgi if you prefer.

Sunny-Maxis:
@botovq 

These were quick and dirty modifications, so we can fix up some commits to repair these complaints. 

For one:

> Is there a way to check this in CI to reduce the risk of accidental breakage?

IRIX is sgi MIPS only. No CI. But again, maintenance of patches is something that me/my employer can handle. They can't break other OSes because __sgi is only defined on IRIX -- that's the only area that preprocessing may intersect.

crypto/compat/arc4random_irix.h
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright (c) 2025 Kazuo Kuroi (kazuo@irixnet.org)
```
botovq: Most of this file is a copy of the other `arc4random_*.h`, so the original authors should be retained:

```suggestion
 * Copyright (c) 1996, David Mazieres <dm@uun.org>
 * Copyright (c) 2008, Damien Miller <djm@openbsd.org>
 * Copyright (c) 2013, Markus Friedl <markus@openbsd.org>
 * Copyright (c) 2014, Theo de Raadt <deraadt@openbsd.org>
 * Copyright (c) 2025 Kazuo Kuroi (kazuo@irixnet.org)
```

botovq (changes requested):
Apart from the fallback code if opening or reading from `dev/urandom` fails, I think this looks basically ok now.

I figured that CI would be difficult, the idea was just that we would notice much faster if we break you.

Sunny-Maxis:
> This isn't about the old `bzero()` vs `memset()` dispute, but rather about [`explicit_bzero()`](https://man.openbsd.org/explicit_bzero), which is provided in the compat code from [our libc](https://github.com/openbsd/src/blob/a37023f05c72a4916de2ba5ed4a0339e2329e8fa/lib/libc/string/explicit_bzero.c). The point of this is that the `memset()` will likely be optimized away by the compiler since it is at the end of the function, but we want to make sure that the secret randomness is cleared.


With that in mind I will fix the function to use explicit_bzero() then.

Sunny-Maxis:
Any updates?

PR #1226, opened 2026-01-22
Enable SHA assembly on macosx-aarch64

4a6f656c:


update.sh
```diff
@@ -170,6 +170,7 @@ echo "LibreSSL version `cat VERSION`"
 echo copying libcrypto source
 rm -f crypto/*.c crypto/*.h
 touch crypto/empty.c
+git checkout crypto/crypto_assembly.h
```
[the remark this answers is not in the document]
4a6f656c: For reasons unknown we remove all `*.h` files two lines earlier... we won't need this once `crypto_assembly.h` is upstream though.

patches/sha-aarch-ce-macos-llvm.patch
```diff
@@ -0,0 +1,299 @@
+Index: crypto/sha/sha1_aarch64_ce.S
```
busterb, line 1: Why not just check in a standalone version of sha1_aarch64_ce_darwin.S at this point? Is it necessary to have all of this indirection for the sake of having one file instead of two? How often is this implementation going to change anyway?

Things only gets worse to impossible with armasm64 from VS, and once we start touching the stack like in sha512 and handling SEH. The complexity of trying to reinvent perlasm in cpp seems it would dwarf any benefits.

update.sh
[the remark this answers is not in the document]
busterb: The `rm` above could be replaced with `git clean -f -- crypto/*.c crypto/*.h` - the original intent was just to delete untracked files that had been copied in from a previous build, but adding a tracked .h file changes that assumption

busterb:
I rebased this on master and applied a few fixes; would you like me to just push to this PR, open a new one, etc.?

This needs a rebase since it no longer works on the latest openbsd, or I'd just PR diffs alone. Changes here: https://github.com/libressl/portable/compare/master...busterb:portable:pr-1226

crypto/CMakeLists.txt
```diff
@@ -69,6 +69,20 @@ if(HOST_ASM_ELF_X86_64)
 	set(CRYPTO_SRC ${CRYPTO_SRC} ${ASM_X86_64_ELF_SRC})
 endif()
 
+if(HOST_ASM_MACOSX_AARCH64)
+	set(CRYPTO_SRC ${CRYPTO_SRC} sha/sha1_amd64.c sha/sha256_amd64.c sha/sha512_amd64.c)
```
busterb, line 73: Wrong architecture here.

patches/sha-aarch-ce-macos-llvm.patch
busterb: I could get mingw64 sort of working with this macro approach, but that's about it.

4a6f656c: I'm not against having separate `sha1_aarch64_ce_darwin.S` and `sha1_aarch64_ce_windows.S` or whatever, when there is significant divergence - there are a few options, including some form of templating/generation that we then use to create the .S and check them in (as opposed to perlasm being very opaque and part of the build process).

This branch was really just a proof-of-concept - for darwin I think the differences are resolvable and I was planning on upstreaming the majority of this diff. The mingw64/windows side is the one that I've not yet looked in to.

Issue #1233, opened 2026-02-19
Unable to build v4.2.1 on macOS


sergio-nsk:
Steps:
```
$ git clone https://github.com/libressl/portable.git -b v4.2.1
$ cd portable
$ ./autogen.sh
pulling upstream openbsd source
Cloning into 'openbsd'...
remote: Enumerating objects: 14001, done.
remote: Counting objects: 100% (14001/14001), done.
remote: Compressing objects: 100% (9023/9023), done.
remote: Total 14001 (delta 5474), reused 11833 (delta 4920), pack-reused 0 (from 0)
Receiving objects: 100% (14001/14001), 25.97 MiB | 12.04 MiB/s, done.
Resolving deltas: 100% (5474/5474), done.
Updating files: 100% (13609/13609), done.
fatal: ambiguous argument 'OPENBSD_7_8': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
This is tag v4.2.1, trying OpenBSD tag libressl-v4.2.1
error: pathspec 'libressl-v4.2.1' did not match any file(s) known to git
No matching OpenBSD tag found trying nearest commit 
libcrypto version 57:2:0
libssl version 60:2:0
libtls version 33:2:0
LibreSSL version 4.2.1
copying libcrypto source
generating mips ASM source for elf
generating mips64 ASM source for elf
generating x86_64 ASM source for elf
generating x86_64 ASM source for macosx
generating x86_64 ASM source for masm
generating x86_64 ASM source for mingw64
copying libtls source
copying nc(1) source
copying ocspcheck(1) source
copying openssl(1) source
copying libssl source
copying tests
patching file 'crypto/arch/aarch64/crypto_arch.h'
1 out of 1 hunks failed--saving rejects to 'crypto/arch/aarch64/crypto_arch.h.rej'
```
This did not happen 2 days ago.
```
fatal: ambiguous argument 'OPENBSD_7_8': unknown revision or path not in the working tree.
error: pathspec 'libressl-v4.2.1' did not match any file(s) known to git
```
and then
```
1 out of 1 hunks failed--saving rejects to 'crypto/arch/aarch64/crypto_arch.h.rej'
```

PR #1234, opened 2026-02-20
[DO NOT MERGE] Add `--unshallow` to git fetch in update.sh

nak3:
When the openbsd repository is cloned with `--depth`, it becomes a shallow clone and lacks sufficient history to resolve commits based on timestamps.

Replace git fetch with git fetch --unshallow to ensure the commit history and branch are available.

Fix https://github.com/libressl/portable/issues/1233


nak3:
I think there are other solutions such as:
- add `--no-single-branch` to git clone.
- drop `--depth=8` from git clone. 
...etc
I will investigate more.

PR #1237, opened 2026-03-01
ci: run tests on macos-26

joshuasing:
Add `macos-26` (Tahoe) runner image to macOS workflow.

Apple is still providing security updates for macOS 14 (Sonoma), however to reduce the number of workflows running, maybe we want to consider removing it or only running `arm64` tests on older versions?

Currently all of the macOS images are Apple Silicon - GitHub now provides `-intel` images if we want to do some tests on older Intel-based hardware as well.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["macos-15", "macos-14"]
+        os: ["macos-26", "macos-15", "macos-14"]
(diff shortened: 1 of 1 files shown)


botovq (approved):
I'm ok with this. I have no informed opinion on dropping 14 or using -intel images. Can you keep an eye on it and do the needed surgery if needed? Right now the CI isn't super active, so I don't believe we're close running into limits.

We probably want to keep an intel mac CI for the foreseeable future to ensure @4a6f656c asm surgery doesn't break things there.

PR #1244, opened 2026-04-12
speed: fix SIGALRM handler not registered on Windows

### LINKED ISSUE #1243: Crash during speed test
I ran LibreSSL v[4.2.1](https://github.com/libressl/portable/releases/tag/v4.2.1) on Windows 7 SP1 x64 as follows
``` console
$ openssl.exe speed chacha20-poly1305
WARNING: can't open config file: C:/Windows/libressl/ssl/openssl.cnf
Doing chacha20 poly1305 for 3s on 16 size blocks:
```
Then it crashed.

 

 _Originally posted in [#125](https://github.com/libressl/portable/issues/125#issuecomment-4205625534)_

nak3:
On Windows, the `sigaction()` call is guarded by `#ifndef _WIN32`, but
there was no corresponding `#else` to register the signal handler via
`speed_signal()`.  This left `speed_alarm_handler` as `NULL`, causing
`speed_timer()` to call through a `NULL` function pointer and crash when
the alarm fired.

Add an `#else` branch that calls `signal(SIGALRM, sig_done)`, which is
remapped to `speed_signal()` by the macro defined in the `_WIN32` section,
so that `sig_done` is properly registered as the alarm handler before
any measurement loop begins.

FIx https://github.com/libressl/portable/issues/1243

### DIFF
--- a/patches/speed.c.patch
+++ b/patches/speed.c.patch
@@ -1,5 +1,5 @@
---- apps/openssl/speed.c.orig	2026-03-18 18:00:10
-+++ apps/openssl/speed.c	2026-03-18 18:00:28
+--- apps/openssl/speed.c.orig	2026-04-12 11:09:28
++++ apps/openssl/speed.c	2026-04-12 11:11:18
 @@ -156,7 +156,16 @@ static void print_result(int alg, int run_no, int coun
  pkey_print_message(const char *str, const char *str2,
      int bits, int sec);
@@ -65,7 +65,7 @@
  
  	if (j == 0) {
  		for (i = 0; i < ALGOR_NUM; i++) {
-@@ -1607,11 +1624,13 @@ speed_main(int argc, char **argv)
+@@ -1607,11 +1624,15 @@ speed_main(int argc, char **argv)
  #define COND	(run && count<0x7fffffff)
  #define COUNT(d) (count)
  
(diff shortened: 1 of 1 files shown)


nak3:
#1243 could also be reproduced on my windows env. And this PR was verified that it does not cause the crash.

```
> .\apps\openssl\Debug\openssl.exe speed chacha20-poly1305
WARNING: can't open config file: C:/Windows/libressl/ssl/openssl.cnf
Doing chacha20 poly1305 for 3s on 16 size blocks: 4152811 chacha20 poly1305 in 2.97s
Doing chacha20 poly1305 for 3s on 64 size blocks: 4165747 chacha20 poly1305 in 3.00s
Doing chacha20 poly1305 for 3s on 256 size blocks: 1994550 chacha20 poly1305 in 3.00s
Doing chacha20 poly1305 for 3s on 1024 size blocks: 644504 chacha20 poly1305 in 3.02s
Doing chacha20 poly1305 for 3s on 8192 size blocks: 89537 chacha20 poly1305 in 2.98s
LibreSSL 4.2.1
built on: date not available
compiler: information not available
The 'numbers' are in 1000s of bytes per second processed.
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes
chacha20 poly1305    22381.47k    88869.27k   170201.60k   218850.85k   245775.78k
```

botovq:
Patch looks good, I'll merge it later. Most likely fallout from trying to adapt to https://github.com/openbsd/src/commit/ee713cad9993da8e2c9da0e398db96b690675c89 (I don't have use for nor access to Windows.)

botovq (approved):

PR #1246, opened 2026-04-12
ci: improve cifuzz workflow

joshuasing:
Update the CIFuzz workflow:
- Pin `google/oss-fuzz` actions to commit hash (https://github.com/google/oss-fuzz/commit/1b73904cfccf3addafb6d599b8e0f32e710790cc, last commit to modify https://github.com/google/oss-fuzz/tree/master/infra/cifuzz/actions)
- Tidy up workflow, restrict token permissions to `security-events: write`
- Upload SARIF output from `run_fuzzers` (matches example here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/)

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,4 +1,4 @@
-name: CIFuzz
+name: "CIFuzz"
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


joshuasing:
The CIFuzz example on https://google.github.io/oss-fuzz/getting-started/continuous-integration/ suggests running on `pull_request` - currently we're using `workflow_dispatch` (manual). Do we want to potentially add a schedule to run this, or was it made manual for a reason?

botovq:
This was changed in 477e8e1f49cc49d22ceae03bad915d6ef5fee994 because it kept failing. Does this work again?

The changes look good at a first glance. I'll review this and #1247 in detail after sleep.

botovq (approved):
I'm gonna land this. We can make further tweaks on top.

PR #1247, opened 2026-04-12
ci: add renovate-bot config

joshuasing:
Add `.github/renovate.json5` configuration file for Renovate automated dependency update PRs.

This configuration will cause Renovate to:
- create pull requests for pending updates on weekends
- pin updated versions
- only update dependencies once the new version has been available for >3 days, lowering risk of using a compromised version and providing time for security scanners/researchers to check updated versions
- rebase stale pull requests
- pin github actions to digests
- use commit message format `ci: update <dep> to v0.0.0`, or `ci: pin <dep> to <digest>`
- add `dependencies` label to pull requests, `security` label when the dependency update contains a security fix
- enable vulnerability alerts, which will create PRs bypassing the weekend schedule

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,32 @@
+{
+  $schema: 'https://docs.renovatebot.com/renovate-schema.json',
+  extends: [
+    'config:best-practices',
+    ':configMigration',
+    ':pinVersions',
+    ':rebaseStalePrs',
+    'schedule:weekends',
+    'customManagers:githubActionsVersions',
+    'helpers:pinGitHubActionDigests',
+  ],
+  commitMessagePrefix: 'ci:',
+  commitMessageAction: 'update',
+  commitMessageTopic: '{{depName}}',
+  labels: [
+    'dependencies',
+  ],
+  minimumReleaseAge: '3 days',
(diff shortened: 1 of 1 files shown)


botovq (approved):

busterb:
What dependencies do we have that need regular updating that renovate would be aware of? Is this just bumping github actions?

botovq:
> What dependencies do we have that need regular updating that renovate would be aware of? Is this just bumping github actions?

Yes. These actions are pinned to a hash because not doing that is dangerous. But bumping them by hadn is a pain (there are tools, but if we can automate it, that's more convenient).

botovq:
@joshuasing Perhaps it's not Monday quite yet where renovate-bot lives, but I would have expected it to have bumped the checkout actions to https://github.com/actions/checkout/commit/de0fac2e4500dabe0009e67214ff5f5447ce83dd aka `v6.0.2` by now, e.g., here

https://github.com/libressl/portable/blob/47003a364c2f62482a0f8e8f9090269937384099/.github/workflows/linux.yml#L47

what's missing?

joshuasing:
It looks like it hasn't picked up the config for some reason... I will have a look. Thanks

PR #1248, opened 2026-04-15
Finish macOS arm64 SHA assembly support

busterb:
This is just PR #1226 rebased on master and with a few additional fixes applied on top.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -123,6 +123,8 @@ AM_CONDITIONAL([HOST_ASM_ELF_MIPS64],
     [test "x$HOST_ABI" = "xelf" -a "$host_cpu" = "mips64" -a "x$enable_asm" != "xno"])
 AM_CONDITIONAL([HOST_ASM_ELF_X86_64],
     [test "x$HOST_ABI" = "xelf" -a "$host_cpu" = "x86_64" -a "x$enable_asm" != "xno"])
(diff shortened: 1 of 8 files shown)


botovq:
I think you'll need an ok from @4a6f656c for this.

Unless I'm missing something, the crypto_assembly.h part clobbers the upstream one. That should probably be a patch instead (too close to OpenBSD full lock).

*Edit:* I'd probably prefer landing this early in the next cycle, but maybe it's trivial enough. I don't know.

busterb:
You're right, there's logic now in update.sh that prevents upstream's minimal crypto_assembly.h from clobbering this one. It could be a patch too, I was just copying the pattern that was in @4a6f656c 's original draft PR. If they don't have time to review, I can just drop this for now.

busterb:
Sounds like this isn't going to happen. I'll just close it.

botovq:
It didn't sound like anything...

I really don't like seeing non-trivial effort go to waste like this. It's very frustrating. If this has no chance of landing, it would be nice if @4a6f656c could say so quickly instead of letting things rot for such a long time. On the other hand, if it does have a chance of landing, it would also be nice to have some sort of feedback in that direction.

4a6f656c:
Apologies for not seeing this earlier. The majority of the changes are now upstreamed (which was always the plan), so the only part needed should be the build glue.

PR #1254, opened 2026-04-19
Allow building on AIX

### LINKED ISSUE #1116: AIX vsyslog
AIX neither has vsyslog nor vsyslog_r.
```
ld: 0711-317 ERROR: Undefined symbol: .vsyslog_r
```
This happens with 4.0.0 as well as 3.9.2 and likely others. Adding the following to crypto/cryptlib.c worked for me:

```
#ifdef _AIX
#ifdef HAVE_SYSLOG
void vsyslog(int facility_priority, const char *format, va_list arglist) {
	char *msg = NULL;
	vasprintf(&msg, format, arglist);

	if (!msg)
		return;

	syslog(facility_priority, "%s", msg);
	free(msg);
}
#endif /* HAVE_SYSLOG */

void vsyslog_r(int pri, struct syslog_data *data, const char *fmt, va_list ap) {
	vsyslog(pri, fmt, ap);
}
#endif /* AIX */
```

botovq:
From @AIIleG in #1116
Fixes #1116

### DIFF
--- a/include/compat/endian.h
+++ b/include/compat/endian.h
@@ -158,4 +158,71 @@
 #define htobe64(x) BE_64(x)
 #endif
 
+#ifdef _AIX /* AIX is always big endian */
+#include <stdint.h>
+#include <sys/types.h>
+
+static inline uint64_t htole64(uint64_t x) {
+#ifdef __BYTE_ORDER__
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+	return x;
+#else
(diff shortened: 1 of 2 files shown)


botovq:
I don't mind carrying these as long as the patches remain small like this. I don't think we have a way to ensure we don't break things down the road so we won't officially support it, but it will likely remove friction for downstreams.

I'd also suggest we backport this so it will be available in 4.3.2.

PR #1269, opened 2026-04-30
Do not override user CFLAGS/LDFLAGS in hardening flag detection

### LINKED ISSUE #1268: Incorrect -fcf-protection detection for GCC 15 and -march=i586 in CFLAGS
When I'm trying to build LibreSSL portable 4.3.1 with GCC 15.2.1 for i586 Linux using autotools, I get the following compilation error:

```
cc1: error: '-fcf-protection' is not compatible with this target
```

I pass `-march=i586` in CFLAGS, but the configure script does not use my CFLAGS when checking for `-fcf-protection` support:

```
$ grep fcf-protection config.log | head -n2
configure:13784: checking if i586-alt-linux-gcc supports "-fcf-protection=full"
configure:13799: ./scripts/wrap-compiler-for-flag-check i586-alt-linux-gcc -o conftest -fcf-protection=full -Wall -Werror  -D_DEFAULT_SOURCE -D_BSD_SOURCE -D_POSIX_SOURCE -D_GNU_SOURCE  conftest.c  >&5
```

And for our i586 GCC default `-march` is i686:

```
$ i586-alt-linux-gcc -Q --help=target | grep '^\s\+-march'
  -march=                               i686
```

GCC 15 does support `-fcf-protection` on i686, so it is marked as supported:

```
checking if i586-alt-linux-gcc supports "-fcf-protection=full"... yes
```

But then, when my CFLAGS are used together with what `configure` detected, the compilation fails.

nak3:
`CHECK_CFLAG` / `CHECK_LDFLAG` overwrote `CFLAGS`/`LDFLAGS` with only the flag under test, so feature detection ran with a different target than the real build.

For example `CFLAGS=-march=i586 ./configure` accepted `-fcf-protection=full` (detected under the toolchain's i686 default) but then failed to compile because GCC 15 rejects `-fcf-protection` on i586.

Prepend `$USER_CFLAGS` / `$USER_LDFLAGS` (saved before `AC_PROG_CC`) to the detection command so the probe matches the real build.

Fixes #1268

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -26,6 +26,7 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
 # This must be saved before AC_PROG_CC
 USER_CFLAGS="$CFLAGS"
+USER_LDFLAGS="$LDFLAGS"
 
 AC_PROG_CC([cc gcc])
 AM_PROG_CC_C_O
--- a/m4/check-hardening-options.m4
+++ b/m4/check-hardening-options.m4
@@ -3,7 +3,7 @@ AC_DEFUN([CHECK_CFLAG], [
 	 AC_LANG_ASSERT(C)
 	 AC_MSG_CHECKING([if $saved_CC supports "$1"])
 	 old_cflags="$CFLAGS"
-	 CFLAGS="$1 -Wall -Werror"
+	 CFLAGS="$USER_CFLAGS $1 -Wall -Werror"
 	 AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <stdio.h>]], [[printf("Hello")]])],
 		     [AC_MSG_RESULT([yes])
 		     CFLAGS=$old_cflags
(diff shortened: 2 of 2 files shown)


nak3:
I was able to reproduce the error on my local Linux.

```
$ ./autogen.sh
$ CFLAGS="-march=i586 -m32" LDFLAGS="-m32" ./configure
$ make -j
Making all in include
  ...
cc1: error: '-fcf-protection' is not compatible with this target
cc1: error: '-fcf-protection' is not compatible with this target
cc1: error: '-fcf-protection' is not compatible with this target
cc1: error: '-fcf-protection' is not compatible with this target
cc1: error: '-fcf-protection' is not compatible with this target
cc1: error: '-fcf-protection' is not compatible with this target
cc1: error: '-fcf-protection' is not compatible with this target
cc1: error: '-fcf-protection' is not compatible with this target
   ...
```

And I confirmed that the error is gone after applying this patch. The build did not fully complete in my environment (it failed later for an unrelated reason), but the `-fcf-protection` error itself was clearly fixed.

botovq (approved):
Makes sense and looks good to me.

PR #1272, opened 2026-05-04
ci: rework release workflow for immutable releases

joshuasing:
Switch the release workflow to create a draft release, and upload the build artifacts to the draft, using the `gh` CLI instead of `softprops/action-gh-release`. This prepares for enabling immutable releases, which require all assets to be uploaded prior to publishing.

The release process now requires a maintainer to manually publish the draft release once all CI artifacts have been uploaded and any remaining assets (e.g. the source tarball, checksums and signatures files) have been attached.

While modifying the release workflow:
 - Tighten workflow permissions by defaulting to no permissions and grant `contents: write` to both jobs (necessary to create the release and upload release assets).
 - Set `persist-credentials: false` on actions/checkout so the GITHUB_TOKEN is not left in .git/config during runs.
 - Set `cancel-in-progress: false` to prevent a second tag push from cancelling an already-running workflow and leaving a dangling release.
 - Move all `${{ ... }}` expressions in run blocks into env to avoid shell interpolation of workflow context values.
 - Add `name` key to every step in build-windows.

Zizmor scan output:
```
> ~/.cargo/bin/zizmor --persona auditor ./.github/workflows/release.yml
 INFO zizmor: ? zizmor v1.24.1
 INFO audit: zizmor: ? completed ./.github/workflows/release.yml
help[misfeature]: usage of GitHub Actions misfeatures
  --> ./.github/workflows/release.yml:73:9
   |
73 |         shell: msys2 {0}
   |         ^^^^^^^^^^^^^^^^ shell defined here
74 |         run: ./autogen.sh
   |         --- uses a non-well-known shell
   |
   = note: audit confidence ? High
   = tip: use a shell that's well-known to GitHub Actions, like 'bash' or 'pwsh'

1 finding: 0 informational, 1 low, 0 medium, 0 high
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,37 +6,45 @@ on:
     tags: [ "v*" ]
 
 concurrency:
-  group: "${{ github.workflow }}-${{ github.event.number || github.ref }}"
-  cancel-in-progress: true
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: false
 
-permissions:
-  contents: write
+permissions: {}
 
 jobs:
   release:
     name: "Release"
     runs-on: "ubuntu-24.04"
-    outputs:
-      upload_url: "${{ steps.create_release.outputs.upload_url }}"
+    permissions:
+      contents: write # Required to create release.
     steps:
       - name: "Checkout repository"
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
 
       - name: "Generate version changelog"
         run: .github/scripts/changelog.sh "$VERSION" > release-changelog.txt
         env:
           VERSION: "${{ github.ref_name }}"
 
-      - name: "Create GitHub release"
-        id: create_release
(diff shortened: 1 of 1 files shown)


joshuasing:
Currently the Windows archive names are in the format: `libressl_vX.X.X_windows_{Windows32,x64,ARM64}.zip`

Do we want to leave this as-is or change it to match the source tarball name of `libressl-vX.X.X.tar.gz` (e.g. `libressl-vX.X.X-windows-arm64.zip`)?

botovq (approved):
I am happy with this and would love to to merge this.

@busterb any concerns if I merge or would you prefer to wait until after the 4.3.2 release (in about 2 weeks)?

botovq:
> Currently the Windows archive names are in the format: libressl_vX.X.X_windows_{Windows32,x64,ARM64}.zip
>
>  Do we want to leave this as-is or change it to match the source tarball name of libressl-vX.X.X.tar.gz (e.g. libressl-vX.X.X-windows-arm64.zip)?

I generally prefer consistency, but I am a bit reluctant to change the file names.

Either we stick with the inconsistency forever, which will continue to be a mild annoyance for some of us, or we cause some downstream friction once, which is going to be annoying for some. I do not know who relies on these archives and how, so I don't know wich of these two has more weight.

PR #1277, opened 2026-05-12
CI: prepare move of COVERITY_SCAN_TOKEN to environment

botovq:
There is no need for this to be a repository secret as only the coverity workflow needs this.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -13,6 +13,9 @@ concurrency:
 jobs:
   scan:
     name: "Scan"
+    environment:
+      name: "coverity"
+      deployment: false
     runs-on: "ubuntu-24.04"
     if: github.repository_owner == 'libressl' # Prevent running on forks
     permissions:


.github/workflows/coverity.yml
```diff
@@ -13,6 +13,7 @@ concurrency:
 jobs:
   scan:
     name: "Scan"
+    environment: "coverity"
```
joshuasing: According to the GitHub docs, we can prevent use of an environment from creating a deployment object by setting `deployment: false`:
```suggestion
    environment:
      name: "coverity"
      deployment: false
```

https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/control-deployments#using-environments-without-deployments
https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/deploy-to-environment (4th point)

joshuasing (approved):
Looks good with one comment.

Until recently, environments were always considered "deployments" and would display as "Deployed to coverity" (or similar). It now seems to be possible to disable that behaviour by setting `deployment: false` on the environment: https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/control-deployments#using-environments-without-deployments

.github/workflows/coverity.yml
botovq: Thanks. should be fixed.

PR #1286, opened 2026-05-26
backport CI changes to OPENBSD_7_8 branch 

busterb:
Same as #1285, this should get the OPENBSD_7_8 branch up to date with CI changes from master. I don't think we needed that rust workaround anymore either.


joshuasing (approved):
Thanks!

PR #1301, opened 2026-06-15
ci: detect Visual Studio CMake generator on Windows

nak3:
This PR stops hard-code the Visual Studio generator in the Windows CI matrix. 
Detect the available Visual Studio generator from `cmake --help` and pass only the generator name to CMake.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -13,7 +13,7 @@ concurrency:
 
 jobs:
   test:
-    name: "${{ matrix.os }}/${{ matrix.arch }} (${{ matrix.generator }}${{ matrix.shared == 'ON' && ', shared' || '' }})"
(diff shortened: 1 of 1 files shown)


nak3:
Current windows CI is failing as https://github.com/libressl/portable/actions/runs/27487805799

.github/workflows/windows.yml
```diff
@@ -50,8 +44,17 @@ jobs:
         run: ./update.sh
 
       - name: "Configure CMake"
-        shell: cmd
-        run: cmake -Bbuild -G "${{ matrix.generator }}" -A ${{ matrix.arch }} -D BUILD_SHARED_LIBS=${{ matrix.shared }} -D CMAKE_INSTALL_PREFIX=../local
+        shell: pwsh
```
nak3, line 47: I can keep using `shell: cmd` but the pwsh command is simpler.

botovq: Seems fine to me

botovq (approved):

PR #1309, opened 2026-06-25
fix out-of-bounds write in getdelim on undersized buffer

sage-mode-hunter:
getdelim only (re)allocates when *buf is NULL or *bufsiz is 0, so a too-small caller-supplied buffer slips through:
- a one-byte heap buffer with *bufsiz == 1 is left untouched
- when the first byte read equals the delimiter, the NUL terminator is written at buf[1], one past the end
- getline() inherits this since it calls getdelim()

Grow the buffer to BUFSIZ whenever it cannot hold a byte plus its terminator. AddressSanitizer flags the write at getdelim.c (heap-buffer-overflow, one byte past a one-byte region) before the change and runs clean after.

### DIFF
--- a/crypto/compat/getdelim.c
+++ b/crypto/compat/getdelim.c
@@ -38,10 +38,18 @@ getdelim(char **buf, size_t *bufsiz, int delimiter, FILE *fp)
 	char *ptr, *eptr;
 
 
-	if (*buf == NULL || *bufsiz == 0) {
-		*bufsiz = BUFSIZ;
-		if ((*buf = malloc(*bufsiz)) == NULL)
+	/*
+	 * Ensure the buffer can hold at least one byte plus the NUL
+	 * terminator before the loop writes to it.  A caller-supplied
+	 * buffer smaller than that is grown rather than overrun.
+	 */
+	if (*buf == NULL || *bufsiz < 2) {
+		char *nbuf;
+		size_t nbufsiz = BUFSIZ;
+		if ((nbuf = realloc(*buf, nbufsiz)) == NULL)
 			return -1;
+		*buf = nbuf;
+		*bufsiz = nbufsiz;
 	}
 
 	for (ptr = *buf, eptr = *buf + *bufsiz;;) {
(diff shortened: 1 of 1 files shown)


crypto/compat/getdelim.c
```diff
@@ -38,10 +38,18 @@ getdelim(char **buf, size_t *bufsiz, int delimiter, FILE *fp)
...
 
 
-	if (*buf == NULL || *bufsiz == 0) {
-		*bufsiz = BUFSIZ;
-		if ((*buf = malloc(*bufsiz)) == NULL)
+	/*
+	 * Ensure the buffer can hold at least one byte plus the NUL
+	 * terminator before the loop writes to it.  A caller-supplied
+	 * buffer smaller than that is grown rather than overrun.
+	 */
```
botovq: I think this comment can be omitted without loss.

crypto/compat/getdelim.c
```diff
@@ -38,10 +38,18 @@ getdelim(char **buf, size_t *bufsiz, int delimiter, FILE *fp)
...
+	if (*buf == NULL || *bufsiz < 2) {
+		char *nbuf;
+		size_t nbufsiz = BUFSIZ;
```
botovq: I think nbuf and nbufsiz should have function scope since they're now used twice, so please move the declarations up to follow `ptr` and `eptr`.

botovq (changes requested):
Thanks, this looks good to me.

getdelim is only used by one regress test, and there NULL is passed in the first time around, so this is not really a problem, but it should be fixed anyway.

This code was taken from netbsd, so it would be nice to push that upstream as well.

crypto/compat/getdelim.c
sage-mode-hunter: Dropped it.

crypto/compat/getdelim.c
sage-mode-hunter: Moved both up next to ptr/eptr and dropped the inner declarations in both branches.

sage-mode-hunter:
Both addressed and pushed. Agreed the regress test never hits it since it passes NULL first, but worth tightening regardless. I'll send the same fix to netbsd upstream.

botovq (approved):
Thanks a lot. I'll merge once CI is green.

PR #1314, opened 2026-07-07
guard getdelim buffer growth against size_t overflow

sage-mode-hunter:
The buffer-growth path in getdelim doubles the size without an overflow check:
- once *bufsiz reaches SIZE_MAX/2 (roughly 2 GiB of input on a 32-bit size_t), `*bufsiz * 2` wraps to a small value or zero
- realloc then hands back an undersized buffer while ptr has already been advanced d bytes in, so the next stored byte lands past the allocation
- distinct from the initial-allocation overflow in #1309; this one is the growth branch (`nbufsiz = *bufsiz * 2`)

Reject the growth with EOVERFLOW before the multiply can wrap, matching the size_t overflow guard already used in dirent_msvc.h.

### DIFF
--- a/crypto/compat/getdelim.c
+++ b/crypto/compat/getdelim.c
@@ -27,6 +27,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <errno.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 
@@ -66,6 +68,10 @@ getdelim(char **buf, size_t *bufsiz, int delimiter, FILE *fp)
 		}
 		if (ptr + 2 >= eptr) {
 			ssize_t d = ptr - *buf;
+			if (*bufsiz > SIZE_MAX / 2) {
+				errno = EOVERFLOW;
+				return -1;
+			}
 			nbufsiz = *bufsiz * 2;
 			if ((nbuf = realloc(*buf, nbufsiz)) == NULL)
 				return -1;


crypto/compat/getdelim.c
```diff
@@ -66,6 +68,10 @@ getdelim(char **buf, size_t *bufsiz, int delimiter, FILE *fp)
 		}
 		if (ptr + 2 >= eptr) {
 			ssize_t d = ptr - *buf;
+			if (*bufsiz > SIZE_MAX / 2) {
+				errno = EOVERFLOW;
```
botovq, line 72: I wonder if this should not be `ENOMEM` instead. [POSIX defines `EOVERFLOW` for getdelim()][1] to mean this specific situation, which we aren't guaranteed to hit at this point:

> The number of bytes to be written into the buffer, including the delimiter character (if encountered), would exceed {SSIZE_MAX}.

[1]: https://pubs.opengroup.org/onlinepubs/9799919799/functions/getdelim.html

botovq (commented):
Thanks. The check makes sense to me and again, this will be more interesting for NetBSD than for us, where this is only used to parse a handful of text files in regress whose total size doesn't exceed a couple of megs.

The diff looks correct to me <s>modulo the errno quibble</s>.

crypto/compat/getdelim.c
botovq: Oh, actually we do. I misread. It's SSIZE_MAX, not SIZE_MAX. So we could change the check to be against `SSIZE_MAX` to be pedantic, but `SIZE_MAX / 2` seems good enough.

botovq (approved):

sage-mode-hunter:
Thanks for merging, and for double-checking the POSIX errno wording. Agreed SIZE_MAX / 2 is close enough to the SSIZE_MAX bound in practice.

PR #1315, opened 2026-07-08
handle GetModuleFileName failure in getprogname

sage-mode-hunter:
GetModuleFileName returns an unsigned DWORD, so the error branch never runs:
- length < 0 is always false, so a failed call is not detected
- that falls through and returns the empty or stale static buffer instead of "?"
- terminate progname at the returned length rather than relying on the trailing static-zero byte


botovq (commented):
Thanks again.

This looks good to me. I'd prefer if people more familiar with Windows took a look.

nak3 (approved):
Looks good to me.

PR #1316, opened 2026-07-09
guard _wopendir against unchecked GetFullPathNameW result

sage-mode-hunter:
_wopendir uses the second GetFullPathNameW return as a write offset without rechecking it against the buffer sized from the first call:
- if the path no longer fits the length passed as nBufferLength, the call returns the (larger) required size and leaves the buffer unwritten
- this can happen when the working directory changes between the two calls, which GetFullPathName does not guard against
- p = patt + n then points past the allocation, so the trailing \* and NUL are written out of bounds

Take the offset only when the result is nonzero and below the buffer length.


nak3 (approved):
Looks good to me.

Issue #1273, opened 2026-05-05
missing and unhooked tests


botovq:
Since test source file names aren't unique in the various subdirectories upstream, the test copying only leaves one of them. For example:

```console
$ find /usr/src/regress -name verify.c
/usr/src/regress/lib/libcrypto/x509/bettertls/verify.c
/usr/src/regress/lib/libcrypto/x509/verify.c
/usr/src/regress/lib/libssl/verify/verify.c
$ find /usr/src/regress -name tlstest.c
/usr/src/regress/lib/libssl/tls/tlstest.c
/usr/src/regress/lib/libtls/tls/tlstest.c
$
```

While we could rename these files to deduplicate, I think these should be namespaced when copying, so we can hook them all up in the portable framework (or actively decide against it).

For the relatively important `x509/verify.c` we're also missing all the `regress/libcrypto/certs/` in `update.sh`, particularly the certs, which should end up in the release tarballs. Bundling the million certs for the bettertls test probably isn't worth it, though.

PR #1317, opened 2026-07-10
hook up x509/verify regress test

nak3:
Copy regress/lib/libcrypto/x509/verify.c as x509_verify.c to avoid the basename collision with x509/bettertls/verify.c when regress tests are flattened into tests.

Also copy the libcrypto cert test data and make-dir-roots.pl, add a wrapper to prepare the CApath roots directory, and register the test for both autotools and CMake.

Refs #1273

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,10 +140,14 @@ tests/test.h
 tests/*test.c
 tests/pbkdf2*
 tests/*.pem
+tests/certs/
+tests/make-dir-roots.pl
 tests/testssl
 tests/*.txt
 tests/compat/*.c
 tests/verify*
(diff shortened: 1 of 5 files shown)


botovq:
Thanks for doing this, this looks great. I will review this carefully tomorrow.

botovq (approved):

botovq:
I missed that this introduces a dependency on perl, which previously wasn't there (it was not needed to build from a tarball and run tests, and we should keep it this way):

https://github.com/libressl/portable/actions/runs/29627135221/job/88033657667#step:4:1561

We should probably see if we can detect perl during configure time and only run this test conditionally.

PR #1318, opened 2026-07-11
nul-terminate the snprintf shim on pre-2015 msvc

sage-mode-hunter:
On pre-2015 MSVC (_MSC_VER < 1900) the shim aliases snprintf to _snprintf:
- _snprintf leaves the buffer unterminated when the output is truncated, unlike C99 snprintf
- code that formats untrusted data (cert fields, host names) into a fixed buffer then reads it back as a string runs off the end
- the wrapper forces termination and preserves the existing return value, so valid inputs behave the same
VS2015+ builds compile out this branch and are untouched.

### DIFF
--- a/include/compat/stdio.h
+++ b/include/compat/stdio.h
@@ -57,7 +57,25 @@ int posix_rename(const char *oldpath, const char *newpath);
 #endif
 
 #if defined(_MSC_VER) && _MSC_VER < 1900
-#define snprintf _snprintf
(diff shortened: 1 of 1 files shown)


botovq (approved):
Thanks. I'm not sure we still compile with old MSVC (e.g., printf specifiers for long long and similar), but as long as we have the logic, it makes sense to fix this.

Will merge soon.

PR #1321, opened 2026-07-14
honor the requested timeout in the win32 poll shim

sage-mode-hunter:
poll() on Windows never waits the full timeout, from two spots in the wait loop:
- an infinite wait (timeout -1) assigns timeout_ms = INFINITE, but timeout_ms is int so it truncates back to -1 and the `timespent_ms < timeout_ms` guard exits after the first ~100ms pass
- a socket-only set leaves num_handles 0, so `WAIT_OBJECT_0 + num_handles - 1` underflows to 0xFFFFFFFF and the stale WAIT_FAILED reads as a signaled handle, breaking on the first iteration

Either way an idle blocking poll returns 0, so s_client/s_server spin instead of waiting.

### DIFF
--- a/apps/openssl/compat/poll_win.c
+++ b/apps/openssl/compat/poll_win.c
@@ -264,6 +264,14 @@ poll(struct pollfd *pfds, nfds_t nfds, int timeout_ms)
 
 	do {
 		TIMEVAL tv;
+
+		/*
+		 * Cap the wait at the time remaining so the final pass
+		 * does not overshoot the requested timeout.
+		 */
+		if (timeout_ms != -1 && timeout_ms - timespent_ms < looptime_ms)
+			looptime_ms = timeout_ms - timespent_ms;
+
 		tv.tv_sec = 0;
 		tv.tv_usec = looptime_ms * 1000;
 		int handle_signaled = 0;
--- a/apps/openssl/compat/poll_win.c
+++ b/apps/openssl/compat/poll_win.c
@@ -166,6 +166,7 @@ poll(struct pollfd *pfds, nfds_t nfds, int timeout_ms)
 	 * select machinery
 	 */
 	fd_set rfds, wfds, efds;
+	fd_set rfds_in, wfds_in, efds_in;
 	int rc;
 	int num_sockets;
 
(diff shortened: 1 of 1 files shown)


nak3:
Thank you. I tested your PR on Windows, but the timeout still does not work correctly when polling an idle socket

I requested a 250 ms timeout, but `poll()` returned `-1` with `errno == EINVAL` after only 109 ms:

<details>
<summary> poll_win_timeout.exe </summary>

```
#include <errno.h>
#include <stdio.h>

#include <winsock2.h>
#include <ws2tcpip.h>

#include <poll.h>

int
main(void)
{
        struct sockaddr_in addr;
        struct pollfd pfd;
        WSADATA wsa;
        SOCKET sock;
        ULONGLONG start, elapsed;
        int rc, saved_errno;

        if (WSAStartup(MAKEWORD(2, 2), &wsa) != 0)
                return 1;

        if ((sock = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP)) ==
            INVALID_SOCKET)
                return 1;

        ZeroMemory(&addr, sizeof(addr));
        addr.sin_family = AF_INET;
        addr.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
        addr.sin_port = 0;

        if (bind(sock, (struct sockaddr *)&addr, sizeof(addr)) ==
            SOCKET_ERROR)
                return 1;

        pfd.fd = sock;
        pfd.events = POLLIN;
        pfd.revents = 0;

        errno = 0;
        start = GetTickCount64();
        rc = poll(&pfd, 1, 250);
        saved_errno = errno;
        elapsed = GetTickCount64() - start;

        printf("rc=%d errno=%d elapsed=%llu revents=0x%x\n",
            rc, saved_errno, (unsigned long long)elapsed, pfd.revents);

        closesocket(sock);
        WSACleanup();

        /*
         * No data arrives, so poll must time out normally.
         * Allow some tolerance, but reject the ~100 ms early return.
         */
        if (rc != 0)
                return 1;
        if (elapsed < 200)
                return 1;

        return 0;
}
````
</details>

**Comment by sage-mode-hunter** [comment]
Good catch, and thanks for running it on real Windows. That EINVAL is a second-order effect the earlier fix exposed: select() strips every not-ready descriptor from the fd_sets, so once the first ~100ms pass times out all three sets are empty. The next pass then calls select() with three empty sets, which Winsock rejects with WSAEINVAL, hence the -1 after ~109ms. Before this PR the loop only ever ran one pass so it never showed.

Pushed b901cbc: keep a pristine copy of the sets and restore them before each select() retry. An idle socket now waits out the full timeout and returns 0. Should pass your poll_win_timeout.exe check now.

**Comment by nak3** [comment]
Thank you. I retested commit b901cbc on Windows. The error is fixed and `poll()` now returns `0`:

```
> poll_win_timeout.exe
rc=0 errno=0 elapsed=313 revents=0x0
```

However, the requested 250 ms timeout still takes approximately 300–330 ms. I also tested `poll(&pfd, 1, 201)`, and it took approximately the same amount of time.
It may be worth limiting the final iteration to the remaining time?

sage-mode-hunter:
Makes sense. The loop always slept in fixed 100ms slices, so 250ms rounded up to three full passes. Pushed 9082d30, which clamps each pass to the time remaining, so 250 now runs 100+100+50 and 201 runs 100+100+1. The infinite case is untouched since the clamp only applies when timeout_ms != -1.

nak3 (approved):
Thank you. Looks good to me.

PR #1322, opened 2026-07-15
restore file offset on error in pread/pwrite shims

sage-mode-hunter:
The win32/mingw pread and pwrite shims (tls/compat) emulate the POSIX calls by saving the current offset, seeking to the target, doing the I/O, then seeking back. POSIX requires both to leave the file offset unchanged. On the I/O-error path that contract is broken:

- when read()/write() returns -1 the function returns immediately, so the seek back to the caller's saved position never runs and the descriptor is left positioned at the pread/pwrite target
- pwrite carries the same defect
- a following read/write on that descriptor then runs at the wrong offset

Save the position once, restore it after the I/O on every path, and keep the read/write errno across the restoring lseek.

### DIFF
--- a/tls/compat/pread.c
+++ b/tls/compat/pread.c
@@ -8,21 +8,25 @@
 
 #define NO_REDEF_POSIX_FUNCTIONS
 
+#include <errno.h>
 #include <unistd.h>
 
 ssize_t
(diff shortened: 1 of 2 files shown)


nak3 (approved):
LGTM

PR #1325, opened 2026-07-18
Run x509_verify only when perl is available

nak3:
The x509_verify wrapper uses make-dir-roots.pl to prepare the hashed CApath roots directory.  This introduced a perl dependency for running the test suite from release tarballs.

Detect perl during configure and CMake configuration, and only register x509_verify when perl is available.

Ref https://github.com/libressl/portable/pull/1317#issuecomment-5010022518

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ include(cmake_export_symbol)
 include(GNUInstallDirs)
 
 enable_testing()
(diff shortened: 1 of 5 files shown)


nak3:
https://github.com/nak3/portable/actions/runs/29633841906 verified that x509_verify skips and CI passed.

botovq (approved):
Thank you. This looks great.

PR #1328, opened 2026-07-19
Run bn_mul_div test

nak3:
Although it contains benchmark code used with `--benchmark`, its normal execution path runs `BN_sqr()` regression checks.


botovq (approved):

botovq:
I deliberately omitted this previously because it seemed more trouble than it's worth:

https://github.com/libressl/portable/actions/runs/29795653511/job/88526337772#step:4:1110

There is some coverage of `BN_sqr()` in `bn_test.c` and in `bn_isqrt.c`.

nak3:
Sorry, I created the fix https://github.com/libressl/portable/pull/1332.
The fix skips Solaris CI only but if it is better to skip `bn_mul_div` on all CI, please let me know.

PR #1332, opened 2026-07-21
Skip bn_mul_div test on Solaris

nak3:
The `bn_mul_div.c` depends on TIMEVAL_TO_TIMESPEC, which is not available on Solaris,
causing the test to fail.
This patch skips `bn_mul_div.c` on Slaris.

Also reuse the `SOLARIS` CMake variable for other SunOS-specific checks.

Ref https://github.com/libressl/portable/pull/1328#issuecomment-5030353169


nak3:
Verified this fix on https://github.com/nak3/portable/actions/runs/29807765903/job/88561800583

botovq (approved):
Thanks. The problem really is that `TIMEVAL_TO_TIMESPEC()` is not standardized and not portable. I think we should avoid it in our tests since a number of our users are on quite exotic platforms and this adds needless friction.

In `bn_shift.c` I worked around this by adding a patch, but that's quite ugly:
https://github.com/libressl/portable/blob/7f0d9138798bc97628d64d2f316b335476ab90f7/patches/bn_shift.patch

I'm inclined to land this PR as it is but we should probably just comment out the test in a followup since testing the squares of 0, 1, and 2 really isn't worth much.

nak3:
That makes sense. I'll handle disabling bn_mul_div in a follow-up.

PR #1334, opened 2026-07-23
release CRYPTO_LOCK_UI on open_console error paths

sage-mode-hunter:
open_console() takes CRYPTO_LOCK_UI, then returns failure on two paths without releasing it:
- GetStdHandle(STD_INPUT_HANDLE) returns NULL (services, GUI apps, detached processes with no console)
- stdin is a char device but GetConsoleMode() fails
- UI_process() returns -1 when the open-session callback fails and never calls close_console, the only unlock, so the lock stays held and later passphrase prompts from other threads block on EnterCriticalSection

Drop the lock on those paths; the success paths still hand it to close_console, so return values are unchanged.

### DIFF
--- a/crypto/compat/ui_openssl_win.c
+++ b/crypto/compat/ui_openssl_win.c
@@ -287,11 +287,15 @@ open_console(UI *ui)
 
 	HANDLE handle = GetStdHandle(STD_INPUT_HANDLE);
 	if (handle != NULL && handle != INVALID_HANDLE_VALUE) {
-		if (GetFileType(handle) == FILE_TYPE_CHAR)
-			return GetConsoleMode(handle, &console_mode);
-		else
+		if (GetFileType(handle) == FILE_TYPE_CHAR) {
+			if (GetConsoleMode(handle, &console_mode))
+				return 1;
+		} else
 			return 1;
 	}
+
+	/* UI_process() skips the close callback when open fails; unlock here. */
+	CRYPTO_w_unlock(CRYPTO_LOCK_UI);
 	return 0;
 }
 


botovq:
I wonder if this should not be fixed in [`UI_process(3)`][1] instead:

```diff
diff --git a/lib/libcrypto/ui/ui_lib.c b/lib/libcrypto/ui/ui_lib.c
index 3cfb3d371af..0dd69358757 100644
--- a/lib/libcrypto/ui/ui_lib.c
+++ b/lib/libcrypto/ui/ui_lib.c
@@ -461,8 +461,10 @@ UI_process(UI *ui)
 {
 	int i, ok = 0;
 
-	if (ui->meth->ui_open_session && !ui->meth->ui_open_session(ui))
-		return -1;
+	if (ui->meth->ui_open_session && !ui->meth->ui_open_session(ui)) {
+		ok = -1;
+		goto err;
+	}
 
 	if (ui->flags & UI_FLAG_PRINT_ERRORS)
 		ERR_print_errors_cb(print_error, ui);

```

This would also fix a similar issue in `ui_openssl.c`'s [`open_console()`][2] if [tcgetattr(3)][3] fails with `errno` different from `ENOTTY` and `EINVAL` (which it can according to the docs).

It would also match what OpenSSL have been doing since about 10 years...

[1]: https://github.com/openbsd/src/blob/5643ba99062665afd5c4ed77cdc5106522de3cc6/lib/libcrypto/ui/ui_lib.c#L464-L465
[2]: https://github.com/openbsd/src/blob/5643ba99062665afd5c4ed77cdc5106522de3cc6/lib/libcrypto/ui/ui_openssl.c#L308-L320
[3]: https://man.openbsd.org/tcgetattr#ERRORS

sage-mode-hunter:
Agreed, UI_process() is the better place for this. The win32 open_console takes the lock before anything that can fail, so with your goto err the close callback balances it on the failure paths too, and it covers the tcgetattr case in ui_openssl.c that this patch doesn't reach. One note: the two fixes can't coexist, since unlocking here plus close_console running would double-release the critical section. Since ui_lib.c comes in from openbsd/src, that change belongs upstream; happy to close this once it lands.

PR #1339, opened 2026-07-26
error out when nc minttl is unsupported in set_common_sockopts

sage-mode-hunter:
nc takes -m minttl on every platform, but set_common_sockopts only enforces it where the socket option exists:
- IP_MINTTL is not defined on macOS, so the IPv4 branch is preprocessed away and -m becomes a no-op
- IPV6_MINHOPCOUNT is missing there too, so the IPv6 branch goes the same way and the whole "if (minttl != -1)" body ends up empty
- the IPV6_TCLASS case a few lines up already reports ENOPROTOOPT, and -S and -V reach usage(1) because their getopt cases are compiled out, so minttl was the one requested option still being dropped without a word

Gave it the same treatment as IPV6_TCLASS. On macOS, nc -m 255 now exits with "set IP min TTL not supported: Protocol not available" rather than running with no hop-limit filter on the socket.

### DIFF
--- a/patches/netcat.c.patch
+++ b/patches/netcat.c.patch
@@ -131,7 +131,7 @@
  	}
  	if (Iflag) {
  		if (setsockopt(s, SOL_SOCKET, SO_RCVBUF,
-@@ -1612,13 +1641,17 @@ set_common_sockopts(int s, int af)
+@@ -1612,13 +1641,27 @@ set_common_sockopts(int s, int af)
  	}
  
  	if (minttl != -1) {
@@ -139,6 +139,11 @@
  		if (af == AF_INET && setsockopt(s, IPPROTO_IP,
  		    IP_MINTTL, &minttl, sizeof(minttl)))
  			err(1, "set IP min TTL");
++#else
++		if (af == AF_INET) {
++			errno = ENOPROTOOPT;
++			err(1, "set IP min TTL not supported");
++		}
 +#endif
  
 -		else if (af == AF_INET6 && setsockopt(s, IPPROTO_IPV6,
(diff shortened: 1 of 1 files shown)


patches/netcat.c.patch
```diff
@@ -131,26 +131,36 @@
...
  	if (minttl != -1) {
 +#ifdef IP_MINTTL
  		if (af == AF_INET && setsockopt(s, IPPROTO_IP,
  		    IP_MINTTL, &minttl, sizeof(minttl)))
  			err(1, "set IP min TTL");
++#else
++		if (af == AF_INET) {
++			errno = ENOPROTOOPT;
++			err(1, "set IP min TTL not supported");
++		}
```
botovq:
```suggestion
+		if (af == AF_INET)
+			errx(1, "set IP min TTL not supported");
```

patches/netcat.c.patch
```diff
@@ -131,26 +131,36 @@
...
 -		else if (af == AF_INET6 && setsockopt(s, IPPROTO_IPV6,
 +#ifdef IPV6_MINHOPCOUNT
 +		if (af == AF_INET6 && setsockopt(s, IPPROTO_IPV6,
  		    IPV6_MINHOPCOUNT, &minttl, sizeof(minttl)))
  			err(1, "set IPv6 min hop count");
++#else
++		if (af == AF_INET6) {
++			errno = ENOPROTOOPT;
++			err(1, "set IPv6 min hop count not supported");
++		}
```
botovq:
```suggestion
+		if (af == AF_INET6)
+			errx(1, "set IPv6 min hop count not supported");
```

botovq (commented):
Thanks, makes sense to me. I don't think setting `errno = ENOPROTOOPT;` adds a lot and if we want to print its error string, we should add [`errc(3)`][1] to our `include/compat/err.h` and use that instead.

[1]: https://man.openbsd.org/errc

sage-mode-hunter:
Agreed, the errno dance wasn't buying anything without errc(3). Switched both branches to errx as suggested and reverified the patch still applies clean to the openbsd tree.

botovq (approved):
Thanks. mips32 looks like a hiccup, I'll kick CI off once more a bit later today.

PR #1341, opened 2026-07-29
add libssl verify test

### LINKED ISSUE #1273: missing and unhooked tests

nak3:
Copy `regress/lib/libssl/verify/verify.c` as `ssl_verify.c` to avoid
basename collisions when regress tests are flattened into tests.

Also document x509/bettertls is not included.

Fix https://github.com/libressl/portable/issues/1273


nak3:
Although I created this PR, I found that `openbsd/src/regress/lib/libssl/verify/create-libressl-test-certs.pl` requires the `IO::Socket::SSL::Utils` to be installed on each platfrom.
It is fine to skip the `libssl/verify/verify.c` without merging this PR.

tests/CMakeLists.txt
```diff
@@ -844,6 +844,28 @@ add_executable(ssl_verify_param ssl_verify_param.c)
 target_link_libraries(ssl_verify_param ${OPENSSL_TEST_LIBS})
 add_platform_test(ssl_verify_param ssl_verify_param)
 
+# ssl_verify
+add_executable(ssl_verify ssl_verify.c)
+target_link_libraries(ssl_verify ${OPENSSL_TEST_LIBS})
+if(EMSCRIPTEN)
+	set_target_properties(ssl_verify PROPERTIES LINK_FLAGS
+	    "-sALLOW_MEMORY_GROWTH -sNODERAWFS=1")
+endif()
```
botovq: I'm not sure what `-sNODERAWFS=1` is doing... Can we use the abstraction instead?
```suggestion
prepare_emscripten_test_target(ssl_verify)
```

botovq (commented):
>  It is fine to skip the libssl/verify/verify.c without merging this PR.

Do you think it adds more complications than it's worth? Since you've already done the work, I think we can merge this, but I'm also fine if we document deliberately skip this test (and bettertls).

tests/CMakeLists.txt
nak3: Ah, yes. Thank you. Updated.

nak3:
> Do you think it adds more complications than it's worth? Since you've already done the work, I think we can merge this, but I'm also fine if we document deliberately skip this test (and bettertls).

Thank you. My only concern was the additional dependency on `IO::Socket::SSL::Utils`. Adding it on each platform required a bit of extra work, but it is working now, so please go ahead and merge it.
If the perl package becomes burdensome to maintain in the future, please feel free to skip it.

botovq (approved):

PR #1342, opened 2026-07-30
verify the accepted peer in the win32 socketpair emulation

sage-mode-hunter:
Two ways to take over the loopback rendezvous the win32 socketpair() uses to pair its halves:
- SO_REUSEADDR does not reserve the port on Windows, it lets another process bind the same address and port
- accept() is trusted to be our own connect(), so whatever reaches the ephemeral port first becomes socket_vector[1]
- either way socketpair() still returns 0, with the two halves connected to different peers

Only tlstest and signertest link this today. Switched the listener to SO_EXCLUSIVEADDRUSE and matched the accepted peer against socket_vector[0] before pairing.

### DIFF
--- a/tests/compat/pipe2.c
+++ b/tests/compat/pipe2.c
@@ -74,8 +74,15 @@ int socketpair(int domain, int type, int protocol, int socket_vector[2])
 		.sin_port = 0,
 	};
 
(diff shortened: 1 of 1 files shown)


nak3 (approved):
Looks good to me.

PR #1343, opened 2026-07-31
hook up the x509 callback tests 

nak3:
Copy callback.pl and add a wrapper that prepares the hashed roots
directory and checks the callback output.

Register callback, callbackfailures, and expirecallback with both
autotools and CMake.


nak3:
https://github.com/nak3/portable/pull/6 verified that this PR and `regress/lib/libcrypto/x509/callbackfailures.c` change passes.

botovq:
I have synced your last commit, so I think you can mark this as ready and kick off CI once more:

https://github.com/libressl/openbsd/commit/c7810630ba9823137bac2dbbb150d43088de37d6

botovq (approved):

nak3:
Thank you! The test was executed and passed.

```
PASS: x509_callbacks.sh
```

PR #1345, opened 2026-07-31
don't override pledge and unveil when the host provides them

sage-mode-hunter:
1. include/compat/unistd.h defines pledge() and unveil() as no-op macros with no guard, so on a host that has them the macro shadows the real declaration #include_next just brought in.
2. every call site in openssl(1), nc(1) and ocspcheck(1) then expands to the constant 0 and the == -1 checks pass, so those programs run unconfined with nothing reported.
3. configure already probed for pledge without ever using the result; added the matching unveil probe and the two cmake checks so both build systems set the guards.

Hosts that lack pledge/unveil keep the no-op macros and are unaffected.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,6 +343,16 @@ if(HAVE_GETPROGNAME)
 	add_definitions(-DHAVE_GETPROGNAME)
 endif()
 
+check_symbol_exists(pledge "unistd.h" HAVE_PLEDGE)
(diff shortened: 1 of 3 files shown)


botovq (approved):
I have checked that the apps indeed run without pledge and unveil on openbsd and that your diff fixes that.

Thanks.

sage-mode-hunter:
thanks for verifying it on openbsd and merging

PR #1347, opened 2026-07-31
cmake: use full sysconf path for OPENSSLDIR

### LINKED ISSUE #1346: CMake sets -DTLS_DEFAULT_CA_FILE=\"etc/ssl/cert.pem\"
Found while looking at #1345.

I have no idea why `CMAKE_INSTALL_SYSCONFDIR` defaults to `etc`, but surely the result that `-DTLS_DEFAULT_CA_FILE` is `etc/ssl/cert.pem` isn't desirable.

```console
$ ./build/apps/nc/nc -czv libressl.org 443
nc: unveil etc/ssl/cert.pem: No such file or directory
```

```
build/build.ninja:  DEFINES = -DHAVE_ARC4RANDOM_BUF -DHAVE_ARC4RANDOM_UNIFORM -DHAVE_ARPA_NAMESER_H -DHAVE_ASPRINTF -DHAVE_ATTRIBUTE__BOUNDED__ -DHAVE_ATTRIBUTE__DEAD__ -DHAVE_CLOCK_GETTIME -DHAVE_ENDIAN_H -DHAVE_ERR_H -DHAVE_EXPLICIT_BZERO -DHAVE_FTRUNCATE -DHAVE_GETDELIM -DHAVE_GETENTROPY -DHAVE_GETLINE -DHAVE_GETOPT -DHAVE_GETPAGESIZE -DHAVE_GETPROGNAME -DHAVE_GNU_STACK -DHAVE_MACHINE_ENDIAN_H -DHAVE_MEMMEM -DHAVE_NETINET_IP_H -DHAVE_PLEDGE -DHAVE_REALLOCARRAY -DHAVE_STRCASECMP -DHAVE_STRLCAT -DHAVE_STRLCPY -DHAVE_STRNDUP -DHAVE_STRNLEN -DHAVE_STRSEP -DHAVE_STRTONUM -DHAVE
_SYSLOG -DHAVE_SYSLOG_R -DHAVE_TIMESPECSUB -DHAVE_TIMINGSAFE_BCMP -DHAVE_TIMINGSAFE_MEMCMP -DHAVE_UNVEIL -DLIBRESSL_INTERNAL -DOPENSSL_NO_HW_PADLOCK -DTLS_DEFAULT_CA_FILE=\"etc/ssl/cert.pem\" -D__BEGIN_HIDDEN_DECLS="" -D__END_HIDDEN_DECLS=""
```

nak3:
`CMAKE_INSTALL_SYSCONFDIR` is relative to the install prefix by default. Using it for `OPENSSLDIR` causes `TLS_DEFAULT_CA_FILE` to be set to the relative path `etc/ssl/cert.pem`.

Use `CMAKE_INSTALL_FULL_SYSCONFDIR` for the runtime path while retaining `CMAKE_INSTALL_SYSCONFDIR` for the install destination.

Fix #1346

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -520,7 +520,7 @@ if(OPENSSLDIR STREQUAL "")
 	if(WIN32)
 		set(OPENSSLDIR "C:/Windows/libressl/ssl")
 	else()
-		set(OPENSSLDIR "${CMAKE_INSTALL_SYSCONFDIR}/ssl")
+		set(OPENSSLDIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/ssl")
 	endif()
 
 	set(CONF_DIR "${CMAKE_INSTALL_SYSCONFDIR}/ssl")


nak3:
Verified with these commands:

**before fix:**

```
$ cmake -S . -B /tmp/libressl-before-build -G Ninja \
  -DLIBRESSL_TESTS=OFF \
  -DCMAKE_INSTALL_PREFIX=/tmp/libressl-before-prefix \
  -DCMAKE_EXPORT_COMPILE_COMMANDS=ON

$ grep -m 1 -o 'TLS_DEFAULT_CA_FILE=[^ ]*' \
  /tmp/libressl-before-build/compile_commands.json
```

**result:**

```
TLS_DEFAULT_CA_FILE=\\\"etc/ssl/cert.pem\\\"
```

**after fix:**

```
$ cmake -S . -B /tmp/libressl-after-build -G Ninja \
  -DLIBRESSL_TESTS=OFF \
  -DCMAKE_INSTALL_PREFIX=/tmp/libressl-after-prefix \
  -DCMAKE_EXPORT_COMPILE_COMMANDS=ON

$ grep -m 1 -o 'TLS_DEFAULT_CA_FILE=[^ ]*' \
  /tmp/libressl-after-build/compile_commands.json
```

**result:**

```
TLS_DEFAULT_CA_FILE=\\\"/tmp/libressl-after-prefix/etc/ssl/cert.pem\\\"
```

CMakeLists.txt
```diff
@@ -520,7 +520,7 @@ if(OPENSSLDIR STREQUAL "")
 	if(WIN32)
 		set(OPENSSLDIR "C:/Windows/libressl/ssl")
 	else()
-		set(OPENSSLDIR "${CMAKE_INSTALL_SYSCONFDIR}/ssl")
+		set(OPENSSLDIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/ssl")
 	endif()
 
 	set(CONF_DIR "${CMAKE_INSTALL_SYSCONFDIR}/ssl")
```
botovq, line 526: Should `CONF_DIR` not be changed accordingly?

```suggestion
	set(CONF_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/ssl")
```

botovq: Or maybe we should do `set(CONF_DIR "${OPENSSLDIR})` unconditionally outside of the if/else construct.

nak3: That would make `CONF_DIR` absolute, so `cmake --install --prefix` would no longer affect these files, reintroducing #1118.

With said, this PR has a limitation: changing the prefix with `cmake --install --prefix` installs the CA files in the new location from configuration time, but libtls still looks for them under the configure time prefix. However, I think it may be acceptable not to support this use case.

botovq: I do not understand how `cmake --install --prefix` could possibly work with a preprocessor macro like `TLS_DEFAULT_CA_FILE`.

None of this makes much sense to me, I'm sorry:
```
install(FILES cert.pem openssl.cnf x509v3.cnf DESTINATION ${CONF_DIR})
```
tells me that the certs end up in `${CONF_DIR}` and that's where `TLS_DEFAULT_CA_FILE` should point at, no?

I think cmake tries to do something that we cannot support.

nak3: > I do not understand how cmake --install --prefix could possibly work
> with a preprocessor macro like TLS_DEFAULT_CA_FILE.

Yes, I agree that an install-time `--prefix` cannot update `TLS_DEFAULT_CA_FILE` after it has been compiled into libtls.
However, #1118 was merged to make the configuration files honor `cmake --install --prefix`, and curl-for-win switched from `DESTDIR` to `--prefix` after that fix:

https://github.com/curl/curl-for-win/commit/947dbc1cdb7c142448381e1fb4797b27c5e907e1

> tells me that the certs end up in ${CONF_DIR} and that's where
> TLS_DEFAULT_CA_FILE should point at, no?

You are right. However, looking at curl-for-win, I found that it removes libtls from the package. `TLS_DEFAULT_CA_FILE` is only used by libtls, so the path mismatch does not affect the curl-for-win.

Nevertheless, the curl-for-win build still runs `cmake --install --prefix`. Making `CONF_DIR` absolute would cause that install step to ignore the requested prefix and effectively reintroducing #1118 and breaking the curl-for-win build.
I understand your point, but we need to consider this existing downstream use of `--prefix`.

botovq: I see. I'm not sure why that use of `--prefix` is useful or necessary, but apparently it's a thing on Windows according to its documentation. As far as I can remember, I always set the prefix at configure time for anything I built, it never even occurred to me that you'd want to change that at install time.

There are various uses of `OPENSSLDIR` in libcrypto, which will become incoherent if you use an install path different from the configured one. I guess that doesn't matter for curl on windows, but we should really nudge users towards using `cmake --install-prefix` at configure time, which does the right thing, as far as I can see.

All that said, I think we should merge this and see if anyone complains.

nak3: > All that said, I think we should merge this and see if anyone complains.

Thank you. Agreed. This PR fixes the normal case where installation uses the same prefix selected at configure time. It does not make libtls relocatable when the prefix is overridden at install time, but that was not a working use case before this change either.

PR #1349, opened 2026-08-03
cmake: use the builtin arc4random on weakly seeded platforms

sage-mode-hunter:
The cmake build has no equivalent of the autotools USE_BUILTIN_ARC4RANDOM checks:
- m4/check-os-options.m4 refuses the platform arc4random on macOS < 10.12, FreeBSD < 12 and NetBSD < 7, where it falls back to a weak seed when /dev/random cannot be opened
- CMakeLists.txt takes whatever check_symbol_exists finds, so compat/arc4random.c is never built on those targets and every arc4random_buf() in libcrypto lands in the libc version configure.ac already rejects
- the macOS probe has to gate HAVE_GETENTROPY too: check_function_exists finds getentropy weakly linked against a current SDK even at -mmacosx-version-min=10.11, which would leave the builtin arc4random calling a symbol that is null at runtime

With -DCMAKE_OSX_DEPLOYMENT_TARGET=10.11 libcrypto.a goes from an undefined _arc4random_buf to defining _libressl_arc4random_buf and _getentropy. Default targets keep the values they had, and ctest is green on both.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,7 @@ endif()
 
 project(LibreSSL LANGUAGES C ASM)
 
+include(CheckCSourceCompiles)
 include(CheckFunctionExists)
 include(CheckSymbolExists)
 include(CheckLibraryExists)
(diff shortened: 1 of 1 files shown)


botovq:
While I'm not against merging this since you've already done the work, I wonder: do we really need to care about such ancient OS releases that have been out of support for at least half a decade?

sage-mode-hunter:
Fair question. For me it was less about those releases and more about the two build systems disagreeing: autotools deliberately refuses the libc arc4random there while cmake silently accepts it, and the failure mode is a weakly seeded RNG in libcrypto with no build-time hint. If you'd rather stop caring about those targets entirely, dropping the checks from m4/check-os-options.m4 instead would resolve the divergence just as well, and I'm fine with that outcome too.

botovq (approved):

PR #1351, opened 2026-08-08
Build Emscripten targets in parallel

nak3:
Pass `--parallel` 4 to `cmake --build` for both the regular and ASAN Emscripten jobs.


nak3:
In my GitHub Actions job for Emscripten, the runtime was reduced from about 25 minutes to around 19 minutes.

botovq (approved):
Great. It's tedious to wait for these jobs, so whatever we can win helps.

PR #1352, opened 2026-08-08
Install CI dependencies in workflows

nak3:
Move APT package installation from scripts/test to the CI workflows.

Build the Linux package list based on the target architecture so that CMake, Ninja, QEMU, and the required cross-compilers are installed together. This makes the packages required by each target explicit and avoids repeated apt-get operations within each CI job.


nak3:
> avoids repeated apt-get operations within each CI job.

This reduced the runtime of each Linux GitHub Actions job by one or two minutes.

botovq (approved):

PR #1353, opened 2026-08-08
fix double free of lock in pthread_mutex_destroy

sage-mode-hunter:
pthread_mutex_lock lazily allocates the CRITICAL_SECTION into mutex->lock (NULL means not yet allocated), and PTHREAD_MUTEX_INITIALIZER starts it NULL. pthread_mutex_destroy accounted for neither:
- destroying a statically-initialized or never-locked mutex passes NULL to DeleteCriticalSection
- the freed pointer was left in place, so a second destroy double-frees it and a lock after destroy reuses freed memory

Guarding on non-NULL and clearing mutex->lock after the free keeps destroy consistent with the lazy-init the lock path already relies on.

### DIFF
--- a/include/compat/pthread.h
+++ b/include/compat/pthread.h
@@ -110,8 +110,11 @@ pthread_mutex_unlock(pthread_mutex_t *mutex)
 static inline int
 pthread_mutex_destroy(pthread_mutex_t *mutex)
 {
-	DeleteCriticalSection(mutex->lock);
-	free(mutex->lock);
+	if (mutex->lock != NULL) {
+		DeleteCriticalSection(mutex->lock);
+		free(mutex->lock);
+		mutex->lock = NULL;
+	}
 	return 0;
 }
 


botovq:
It makes sense to do this, because the lazy initialization path is iffy because of the semantics of DeleteCriticalSection.

It is however undefined behavior to call `pthread_mutex_destroy()` on an already destroyed mutex:

> The pthread_mutex_destroy() function shall destroy the mutex object referenced by mutex; the mutex object becomes, in effect, uninitialized

and

> The behavior is undefined if the value specified by the mutex argument to pthread_mutex_destroy() does not refer to an initialized mutex.

So I'm less concerned about that part. See https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_destroy.html

(I'm a bit behind with merging, I'll take care of this in a day or two, sorry about that).

sage-mode-hunter:
Agreed, the double-destroy case is UB so nothing's owed there. The part that matters is destroying a statically-initialized or never-locked mutex, which is legal and hit DeleteCriticalSection(NULL) before; clearing the pointer after free is just cheap hygiene on top. No rush on the merge.

botovq (approved):

PR #1355, opened 2026-08-09
ci: run Windows ARM64 tests on native runners

nak3:
Currently, ARM64 builds are cross-compiled on x64 Windows runners and the tests are skipped.

With this change, static and shared ARM64 builds run on windows-11-arm runners, allowing the test suite to execute natively.

Running the tests natively also exposes incorrect code generation by the MSVC ARM64 optimizer when the generic bn_ct_ne_zero() helper is inlined. Work around this by using the _CountLeadingZeros64() intrinsic on MSVC ARM64, avoiding the affected optimizer transformation and allowing the tests to pass.

### DIFF
--- a/patches/win32_arm64_bn_arch.h.patch
+++ b/patches/win32_arm64_bn_arch.h.patch
@@ -1,0 +1,21 @@
+--- crypto/bn/arch/aarch64/bn_arch.h.orig
++++ crypto/bn/arch/aarch64/bn_arch.h
+@@ -20,4 +20,17 @@
+ #ifndef HEADER_BN_ARCH_H
+ #define HEADER_BN_ARCH_H
+-
++#if defined(_MSC_VER) && defined(_M_ARM64)
++
++#include <intrin.h>
++
++#define HAVE_BN_CT_NE_ZERO
++
++static inline int
++bn_ct_ne_zero(BN_ULONG w)
++{
++	return 1U ^ (_CountLeadingZeros64((unsigned __int64)w) >> 6);
++}
++
++#endif
++
+ #ifndef OPENSSL_NO_ASM
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -22,8 +22,15 @@ jobs:
       fail-fast: false
       matrix:
         os: ["windows-2022", "windows-2025"]
-        arch: ["ARM64", "x64", "Win32"]
+        arch: ["x64", "Win32"]
         shared: ["ON", "OFF"]
+        include:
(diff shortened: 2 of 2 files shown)


nak3:
> Running the tests natively also exposes incorrect code generation by the MSVC ARM64 optimizer when the generic bn_ct_ne_zero() helper is inlined. Work around this by using the _CountLeadingZeros64() intrinsic on MSVC ARM64, avoiding the affected optimizer transformation and allowing the tests to pass.

The reproducer and the report to Microsoft are here:

- Reproducer: https://github.com/nak3/msvc-arm64-bn-repro
- Report https://developercommunity.visualstudio.com/t/MSVC-ARM64-optimizer-emits-ASR-instead-o/11132989

botovq (approved):

patches/win32_arm64_bn_arch.h.patch
```diff
@@ -0,0 +1,21 @@
...
++#if defined(_MSC_VER) && defined(_M_ARM64)
++
++#include <intrin.h>
++
++#define HAVE_BN_CT_NE_ZERO
++
++static inline int
++bn_ct_ne_zero(BN_ULONG w)
++{
++	return 1U ^ (_CountLeadingZeros64((unsigned __int64)w) >> 6);
```
4a6f656c, line 16: Is `_CountLeadingZeros64` guaranteed to be constant time (on arm64 it should produce a single `clz` instruction, but would be good to confirm)?

nak3: Thank you. As you expected, I confirmed that the inlined `_CountLeadingZeros64()` call is lowered to a `CLZ` instruction:

```asm
; 49   :  return 1U ^ (_CountLeadingZeros64((unsigned __int64)w) >> 6);

  00030  dac01108  clz   x8,x8
  00034  d3407d08  uxtw  x8,w8
  00038  d27a0108  eor   x8,x8,#0x40
  0003c  d346fd08  lsr   x8,x8,#6
```

However, as your question implies, producing a `CLZ` instruction does not guarantee constant-time execution. I learned that the Arm architecture guarantees data-independent timing for `CLZ` only when `PSTATE.DIT` is enabled. On the Windows ARM64 runner I tested, the initial value of `PSTATE.DIT` was 0.
Ideally, Windows would enable DIT by default. (like OpenBSD enables DIT for user space by default.)

Having LibreSSL enable and disable DIT around individual low-level operations would be undesirable because DIT is thread-local state owned by the caller, and changing it for every operation would also add overhead.

PR #1357, opened 2026-08-10
strip posix_open tag bit in ftruncate shim

sage-mode-hunter:
ftruncate is the only fd shim that hands its descriptor to the CRT without stripping the posix_open() tag:
- posix_open() sets the 0x80000000 bit, so a descriptor from the wrapped open() is a negative int
- fstat, close, read and write all run it through get_real_fd() first; ftruncate does not
- _chsize() then rejects the negative fd (EBADF) and the truncation silently fails

Move the shim into posix_win.c so it can use get_real_fd() like the others.


botovq:
Can we merge `ftruncate.c` into `posix_win.c`? Then we could use `get_real_fd()` instead of reimplementing it with a long comment. Might need some `HAVE_FTRUNCATE` guards, I don't know.

sage-mode-hunter:
Done. Moved the shim into posix_win.c behind #ifndef HAVE_FTRUNCATE so it just calls get_real_fd(), and dropped ftruncate.c along with its build glue (the AM_CONDITIONAL was only used for that file, so it's gone too). The guard works in both builds since cmake adds -DHAVE_FTRUNCATE and autoconf gets it from AC_CHECK_FUNCS via DEFS. Verified with a mingw cross build that the function is only compiled in when HAVE_FTRUNCATE is unset.

botovq:
Thanks. I think this is better.

Is `HAVE_FTRUNCATE` ever defined if we build `posix_win.c` or can we remove the ftruncate dances from `crypto/CMakeLists.txt` as well?

It feels a bit inconsistent to check for ftruncate only in cmake builds. I guess the question is whether we should keep the `m4/check-libc.m4` bit you remove for symmetry with cmake or if we can simplify further.

sage-mode-hunter:
Yes, mingw-w64 defines it: its unistd.h declares ftruncate (with a comment saying it's a real library function precisely so autoconf checks can find it) and it links from libmingwex, so the check comes back positive there. MSVC has neither unistd.h nor the function, so the check fails and the shim gets compiled. The guard is doing real work and the top-level cmake check has to stay.

On symmetry: autoconf still checks for ftruncate, that's the AC_CHECK_FUNCS line in check-libc.m4 which puts HAVE_FTRUNCATE into DEFS. The bit I removed was only the AM_CONDITIONAL, whose sole consumer was the conditional compile of the deleted file. So both build systems already define HAVE_FTRUNCATE the same way and there's nothing cmake-only left.

I did try simplifying further by dropping the guard and compiling the shim unconditionally. It builds on mingw and the object overrides libmingwex at link time, but that trades mingw's own working ftruncate for a _chsize-backed one, and if anyone ever builds with _FILE_OFFSET_BITS=64 the definition silently gets asm-renamed to ftruncate64, shadowing the real 64-bit version with a 32-bit long backend. Doesn't seem worth it to me.

botovq:
> autoconf still checks for ftruncate, that's the AC_CHECK_FUNCS line in check-libc.m4 which puts HAVE_FTRUNCATE into DEFS.

Oh, right. I missed that.

Rest makes sense, too. I'll merge a bit later. Thanks!

botovq (approved):

PR #1359, opened 2026-08-11
cmake: detect the MSVC ARM64 optimizer bug

nak3:
Replace the compiler version check with a configure-time reproducer.
Native MSVC ARM64 builds use the workaround when the check fails and
the generic implementation when it passes. Cross-compiled builds retain
the workaround because the check cannot be run.

Ref #1355


nak3:
> The issue is fixed in Visual Studio 2026 v18.8.2. Switch the ARM64 jobs to windows-11-vs2026-arm and remove the workaround.

As per the issue report for Microsoft https://developercommunity.visualstudio.com/t/MSVC-ARM64-optimizer-emits-ASR-instead-o/11132989, VS 2026 v18.8.2 fixed the issue and I realized that windows-11-vs2026-arm runner is available.

botovq:
Do we not need to detect this miscompilation problem and use the workaround if it is present?

nak3:
Good point, thank you. I retained the workaround for MSVC versions older than 19.51.36252, the version verified without the workaround on GitHub Actions.

patches/win32_arm64_bn_arch.h.patch
```diff
@@ -1,10 +1,11 @@
 --- crypto/bn/arch/aarch64/bn_arch.h.orig
 +++ crypto/bn/arch/aarch64/bn_arch.h
-@@ -20,4 +20,17 @@
+@@ -20,4 +20,18 @@
  #ifndef HEADER_BN_ARCH_H
  #define HEADER_BN_ARCH_H
 -
-+#if defined(_MSC_VER) && defined(_M_ARM64)
++/* See pull/1355. */
++#if defined(_MSC_VER) && defined(_M_ARM64) && _MSC_FULL_VER < 195136252
```
botovq: I can live with this, but generally I think version checks in the preprocessor are a bad idea.

Ideally, we'd be able to detect the problem using your reproducer if we're on MSVC+arm64 and not cross compiling and enable this codepath in case of failure.

nak3: I see and agreed. I replaced the version check with a runtime configure test.

botovq: Thanks a lot. I like this a lot better.

It seems impossible to see the result of this check in the logs, though or am I being blind?

nak3: I realized that the log is not shown unless it is logged explicitly. The message `MSVC ARM64 optimizer check passed` now appears in the `Configure CMake` step.

Unfortunately, I also found that the detection code in the previous commit was insufficient, and that a long reproducer needs to be added to `CMakeLists.txt`.
Would this still be acceptable?

.github/workflows/windows.yml
```diff
@@ -25,10 +25,10 @@ jobs:
         arch: ["x64", "Win32"]
         shared: ["ON", "OFF"]
         include:
-          - os: "windows-11-arm"
+          - os: "windows-11-vs2026-arm"
```
botovq: Would you mind splitting this (and the change a few lines down) into a separate PR so that we can see the detection work when it's needed?

patches/win32_arm64_bn_arch.h.patch
botovq: > The message MSVC ARM64 optimizer check passed now appears in the Configure CMake step.

Thanks, that's great.

>  Would this still be acceptable?

I think it is fine. If someone is bothered by the length, we can always split this into a separate file and perhaps even check in the reproducer as a .c file.

botovq (approved):

.github/workflows/windows.yml
botovq: I now saw:
```
-- MSVC ARM64 optimizer check failed; workaround enabled (compiled=TRUE, result=1)
```
Awesome. Thank you!

PR #1360, opened 2026-08-12
update.sh: do not ignore git failures

nak3:
git describe is expected to fail when HEAD is not exactly at a tag. However, the set `+e` used for this check also allowed fetch, checkout, and pull failures to be ignored, causing less useful missing-file errors later in the script.

Re-enable set `-e` immediately after recording the git describe result so that subsequent Git operations fail at the actual error.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -30,6 +30,7 @@ fi
 set +e
 tag=`git describe --exact-match --tags HEAD 2>/dev/null`
 is_tag=$?
+set -e
 # adjust for 9 hour time delta between trees
 release_ts=$((`git show -s --format=%ct $tag|tail -1` + 32400))
 commit=`git -C openbsd rev-list -n 1 --before=$release_ts origin/$openbsd_branch`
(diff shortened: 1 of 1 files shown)


nak3:
Please see the following CI run (without this change):

https://github.com/libressl/portable/actions/runs/31547315042/job/93962472440

In this run, a git operation failed because `openbsd/.git/index.lock` already existed. However, since update.sh was still running with set `+e`, the failure was ignored and the script continued:

```
pulling upstream openbsd source
Cloning into 'openbsd'...
   ...
Updating files:  98% (13297/13568)
Not on a tag, grabbing latest (NOTE: this may be broken from time to time)
Updating files:  99% (13433/13568)
fatal: Unable to create '/c/a/portable/portable/openbsd/.git/index.lock': File exists.

Another git process seems to be running in this repository, or the lock file may be stale
  ...
copying ocspcheck(1) source
cp: cannot stat '/c/a/portable/portable/openbsd/src/usr.sbin/ocspcheck/ocspcheck.8': No such file or directory
Error: Process completed with exit code 1.
```

In this case, the later cp command failed and stopped the job. Had it not failed, the CI job could have continued using an incomplete source tree.

botovq (approved):

PR #1363, opened 2026-08-13
ci: run CTest in parallel

nak3:
Run CTest with parallel jobs in the Windows, Emscripten, Fedora, and shared test-script paths to reduce test time.

### DIFF
--- a/.github/workflows/fedora-rawhide.yml
+++ b/.github/workflows/fedora-rawhide.yml
@@ -37,4 +37,4 @@ jobs:
          ninja
     - name: Test
       run: |
-         ninja test
+         CTEST_PARALLEL_LEVEL="" ninja test
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -43,7 +43,7 @@ jobs:
         run: cmake --build build --config Release --parallel 4
 
       - name: "Test"
-        run: ctest --test-dir build -C Release --output-on-failure
+        run: ctest --test-dir build -C Release --parallel --output-on-failure
 
   # Test ASAN with and without ASM enabled.
   test-asan:
@@ -80,4 +80,4 @@ jobs:
         run: cmake --build build --config Release --parallel 4
 
       - name: "Test"
-        run: ctest --test-dir build -C Release --output-on-failure
+        run: ctest --test-dir build -C Release --parallel --output-on-failure
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -88,7 +88,7 @@ jobs:
 
       - name: "Test"
         shell: cmd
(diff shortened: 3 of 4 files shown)


nak3:
One downside is that the test output becomes slightly harder to read. However, I think the faster test execution is worth the trade-off. - [log](https://github.com/nak3/portable/actions/runs/31675399376/job/94368825853)
```
Run ctest --test-dir build -C Release --parallel --output-on-failure
Test project D:/a/portable/portable/build
        Start   1: aeadtest
        Start   2: aes_test
        Start   3: aes_wrap
        Start   4: apitest
  1/140 Test   #3: aes_wrap .........................   Passed    0.09 sec
  2/140 Test   #2: aes_test .........................   Passed    0.09 sec
  3/140 Test   #1: aeadtest .........................   Passed    0.10 sec
        Start   5: asn1api
        Start   6: asn1basic
        Start   7: asn1complex
  4/140 Test   #5: asn1api ..........................   Passed    0.03 sec
        Start   8: asn1evp
  5/140 Test   #6: asn1basic ........................   Passed    0.03 sec
        Start   9: asn1object
  6/140 Test   #7: asn1complex ......................   Passed    0.03 sec
        Start  10: asn1oct
  7/140 Test   #8: asn1evp ..........................   Passed    0.04 sec
        Start  11: asn1string_copy
  8/140 Test   #9: asn1object .......................   Passed    0.04 sec
  9/140 Test  #10: asn1oct ..........................   Passed    0.03 sec
  
   ... continue ...
```

.github/workflows/fedora-rawhide.yml
```diff
@@ -37,4 +37,4 @@ jobs:
          ninja
     - name: Test
       run: |
-         ninja test
+         CTEST_PARALLEL_LEVEL="" ninja test
```
botovq, line 40: I don't mind adding it, but is this bit needed? [ninja documents][1]:

> Note that Ninja defaults to running commands in parallel anyway, so typically you don’t need to pass -j.

[1]: https://ninja-build.org/manual.html#_running_ninja

botovq (approved):
> One downside is that the test output becomes slightly harder to read.

I don't mind at all. This is just for our own purposes and I agree that the faster CI is done the better.

.github/workflows/fedora-rawhide.yml
nak3: Yes, it is needed. ninja runs commands in parallel, but CTest is only one command from ninja. `CTEST_PARALLEL_LEVEL` enables parallelism within CTest itself.

I verified that `CTEST_PARALLEL_LEVEL` works for ninja as:

- before this PR
  - https://github.com/libressl/portable/actions/runs/31660039594/job/94322763682

- after this PR
  - https://github.com/nak3/portable/actions/runs/31751488940/job/94617990485

PR #1364, opened 2026-08-14
zizmor: set persist-credentials: false

botovq:
Per zizmor, actions/checkout should be followed by removal of credentials that
may have been saved to disk. Doing so fixes lots of whining such as:

```
warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> ./.github/workflows/android.yml:27:9
   |
27 |         - name: "Checkout repository"
   |  _________^
28 | |         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
   | |________________________________________________________________________________^ does not set persist-credentials: false
   |
   = note: audit confidence → Low
   = note: this finding has an auto-fix
```

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -26,6 +26,8 @@ jobs:
     steps:
       - name: "Checkout repository"
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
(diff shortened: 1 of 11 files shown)


joshuasing (approved):
Looks good! I have some more zizmor fixes and tidy up locally I'll push up over the weekend

PR #1366, opened 2026-08-16
cmake: apply the same exploit-mitigation flags as the autotools build

sage-mode-hunter:
The autotools build turns on a hardening set by default in CHECK_C_HARDENING_OPTIONS (m4/check-hardening-options.m4), but the CMake build applies none of it, so a cmake-built libcrypto/libssl/libtls and openssl(1) ship without those mitigations:
- no -fstack-protector-strong, no _FORTIFY_SOURCE, no -fno-strict-overflow
- no -Wl,-z,relro / -Wl,-z,now, no -fcf-protection
- mingw DLLs miss the --nxcompat / --dynamicbase / --high-entropy-va opt-ins

This adds an ENABLE_HARDENING block (default on, matching --disable-hardening) that probes each flag before using it, so unsupported targets skip it. Platform gating follows the m4: FORTIFY off on Windows, CET off on Darwin, libssp opt-in behind ENABLE_WINDOWS_SSP. MSVC keeps its own toolchain defaults, same as the m4 which only drives gcc/clang/mingw flags.

Checked on macos/arm64 clang: configure applies -fstack-protector-strong, -fno-strict-overflow and _FORTIFY_SOURCE=2, skips relro/now (Apple ld) and CET, and libcrypto builds clean. -fstack-protector-strong measurably covers functions the default -fstack-protector leaves alone.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,9 +212,13 @@ int main(void){char b[8];const char *s=\"x\";strcpy(b,s);return b[0];}"
 			    "$<$<AND:$<COMPILE_LANGUAGE:C>,$<NOT:$<CONFIG:Debug>>>:-D_FORTIFY_SOURCE=2>")
 		endif()
 
-		# Read-only relocations, resolved at load time.
-		add_harden_ldflag(-Wl,-z,relro HAVE_LDFLAG_Z_RELRO)
-		add_harden_ldflag(-Wl,-z,now HAVE_LDFLAG_Z_NOW)
+		# Read-only relocations, resolved at load time. wasm has no
+		# ELF loader and wasm-ld only warns about unknown -z values,
+		# so the probe false-positives on Emscripten; skip it there.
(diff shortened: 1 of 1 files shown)


botovq:
Doing this makes a lot of sense to me. Some logic for Emscripten seems to be missing. Didn't really have time to think about what's needed.

sage-mode-hunter:
Good catch, the Emscripten CI runs confirmed it. Two things were wrong there: wasm-ld only warns about unknown -z values, so the relro/now probes false-positived and then warned on every link, and the stack protector probe passed (no protected frame in the probe program) while real objects failed to link on the missing __stack_chk_guard/__stack_chk_fail, since Emscripten has its own stack checks instead. Both are now skipped on Emscripten; -fcf-protection already failed its probe cleanly there. What's left on wasm is -fno-strict-overflow and the FORTIFY define, which is where the autotools checks would land too.

botovq (approved):

PR #1370, opened 2026-08-22
Enable supported MIPS32 assembly in CI

nak3:
Use the little endian MIPS32 toolchain since the generated assembly is
little-endian, and define the architecture feature macros needed to
select the AES, BN, SHA-1, and SHA-256 assembly implementations.

Do not generate or build SHA-512 assembly for MIPS32. The generator
accepts the O32 ABI, but the resulting implementation still uses 64-bit
instructions and requires the MIPS64 ISA. It is therefore not valid for
generic MIPS32 targets.


nak3:
This PR requires https://github.com/libressl/portable/pull/1369

> Do not generate or build SHA-512 assembly for MIPS32. The generator
accepts the O32 ABI, but the resulting implementation still uses 64-bit
instructions and requires the MIPS64 ISA. 

The generator script documents this limitation: `SHA512 [which for now can only be compiled for MIPS64 ISA]`
https://github.com/libressl/openbsd/blob/master/src/lib/libcrypto/sha/asm/sha512-mips.pl#L15-L16

botovq (approved):

PR #1371, opened 2026-08-24
clamp read and write length to INT_MAX in win32 posix shims

sage-mode-hunter:
posix_read/posix_write are the process-wide read()/write() on Windows (win32netcompat.h maps read/write to them) and pass their size_t count straight into the int-length socket and CRT calls:
- recv/send take an int length, so a count at or past 2 GiB is narrowed at the call
- count == 4 GiB truncates to 0, so recv() returns 0 and the read loop reads it as a clean EOF
- 2 GiB truncates to INT_MIN and fails a valid buffer with WSAEFAULT; 4 GiB + 1024 sends 1024 and reports it as complete
Clamp count to INT_MAX in both shims before the call. POSIX allows read/write to move fewer bytes than requested so callers loop for the rest, and normal-sized I/O is unchanged.

### DIFF
--- a/crypto/compat/posix_win.c
+++ b/crypto/compat/posix_win.c
@@ -16,6 +16,7 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <limits.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -221,6 +222,8 @@ ssize_t
 posix_read(int fd, void *buf, size_t count)
 {
 	ssize_t rc;
+	if (count > INT_MAX)
+		count = INT_MAX;
 	if (is_socket(fd)) {
 		if ((rc = recv(fd, buf, count, 0)) == SOCKET_ERROR) {
 			int err = WSAGetLastError();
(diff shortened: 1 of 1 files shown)


nak3:
I think it is fine to have this check, but can read() or write() calls exceeding 2 GiB actually occur in LibreSSL?

sage-mode-hunter:
Not with the current callers, as far as I can trace. On the socket side everything goes through the BIO methods (bss_fd/bss_sock/bss_conn/bss_dgram), which all take an int length, and tls_read/tls_write reject buflen > INT_MAX before it gets anywhere near the shim. The only size_t paths are the whole-file reads in tls_load_file() and tls_config_load_file(), which pass st_size straight through, so a cert/CA/CRL file over 2 GiB would reach it. But that fails either way (short read, n != size); the clamp just makes it an ordinary short read instead of a narrowed length. The session pread already caps st_size at INT_MAX, and nc/ocspcheck/s_client use small fixed buffers.

So this is hardening of the shim's size_t contract rather than something a caller hits today, and the PR description reads stronger than that. Fine with keeping it as a cheap guard (I can trim the description to say so), or dropping it if you'd rather not carry it. Your call.

botovq (approved):
I think it is an improvement over the current code and it generally makes sense to handle truncation from wider types to ints.

PR #1378, opened 2026-09-04
fix overwrite guard in the openssldir install hook

sage-mode-hunter:
**Install hook overwrites existing openssldir files**
The guard in install-exec-hook tests `-f "$$OPENSSLDIR/$i"` with a single dollar on `i`, so make expands it as an empty make variable and the shell checks the directory itself, which never passes `-f`. A repeat `make install` then replaces an existing cert.pem, openssl.cnf and x509v3.cnf with the shipped defaults instead of taking the "already exists, install will not overwrite" branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FF1dHqnUEhmE44ucDtFmB2

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -40,7 +40,7 @@ install-exec-hook:
 	fi; \
 	mkdir -p "$$OPENSSLDIR/certs"; \
 	for i in cert.pem openssl.cnf x509v3.cnf; do \
-		if [ ! -f "$$OPENSSLDIR/$i" ]; then \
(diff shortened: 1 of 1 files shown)


botovq (approved):
Leveraging Claude for a one letter fix. What has our world come to? :)

Thanks!

sage-mode-hunter:
Thanks for landing this, the one letter ones tend to hide the longest.

PR #1388, opened 2026-09-16
ci: remove enscripten asan job

joshuasing:
Remove Emscripten ASAN job from CI. It is very slow and has not found anything.


botovq (approved):
good riddance

PR #1393, opened 2026-09-19
add a hook for inserting a release candidate version

busterb:
This adds a hook to the release script so we can create an 'rc' package by specifying LIBRESSL_RC like so:

```
 % LIBRESSL_RC=1 ./dist.sh
pulling upstream openbsd source
Not on a tag, grabbing latest (NOTE: this may be broken from time to time) Already on 'master'
Your branch is up to date with 'origin/master'.
Already up to date.
libcrypto version 57:2:0
libssl version 60:2:0
libtls version 33:2:0
LibreSSL version 4.3.1-rc1
```

Using this should reduce persistent confusion over what the current stable release is the next time the project wants to do a prerelease for testing. Tagging @joshuasing and @botovq to help see if having a tarball called `libressl-4.3.1-rc1.tar.gz` doesn't blow up something in a release process.


joshuasing (approved):

joshuasing:
This shouldn't cause any issues with CI - the rc release process on GitHub would be the same, pushing a tag and then manually publishing the release with uploaded assets.

I much prefer this to the previous confusing method as well. Thanks!

joshuasing:
Probably best to wait for tb's ok before merging, as I am not familiar with other parts of the release process.

update.sh
```diff
@@ -164,6 +164,9 @@ else
 fi
 
 awk '/LIBRESSL_VERSION_TEXT/ {print $4}' < include/openssl/opensslv.h | cut -d\" -f1 | head -n1 > VERSION
+if [ -n "$LIBRESSL_RC" ]; then
+	printf '%s-rc%s\n' "`cat VERSION`" "$LIBRESSL_RC" > VERSION
```
botovq: I would omit the `-`:

```suggestion
	printf '%src%s\n' "`cat VERSION`" "$LIBRESSL_RC" > VERSION
```

botovq (changes requested):
I have absolutely no idea what could possibly blow up. Surely the immutable release stuff on github that is largely outside of our control and the release script for generating that are the most likely to hiccup.

What little I have to do for an actual release definitely doesn't care about inserting "rc1" or "-rc1".

I never understood the motivation for the terrible versioning scheme we've been using and I think releasing one or two release candidates before we commit to the actual release will be a lot clearer for everybody. The last prereleases were treated as stable releases as far as I could tell.

botovq:
What do we do with `LIBRESSL_VERSION_NUMBER` and `LIBRESSL_VERSION_TEXT`? Do we just announce the upcoming `rc` as `4.4.0` or should that include the "rc" in the `_TEXT`? The `_NUMBER` has a weird mechanism that we've promised not to use since forever (see [the manpage](https://man.openbsd.org/OPENSSL_VERSION_NUMBER)):

> `OPENSSL_VERSION_NUMBER` and `LIBRESSL_VERSION_NUMBER` are numeric release version identifiers. The first two digits contain the major release number, the third and fourth digits the minor release number, and the fifth and sixth digits the fix release number. For OpenSSL, the seventh and eight digits contain the patch release number and the final digit is `0` for development, `1` to `e` for betas `1` to `14`, or `f` for release. For LibreSSL, `OPENSSL_VERSION_NUMBER` is always `0x020000000`, and `LIBRESSL_VERSION_NUMBER` always ends with `00f`.

I'd be inclined to leave `LIBRESSL_VERSION_NUMBER` as `0x4040000fL` the same way as in the actual release and consider modifying the `_TEXT` with a patch (this might require moving this awk line to after the patching if that's possible). This way things that really feel they must support and special case an rc can inspect the `_TEXT`. But that might break `FindLibreSSL.cmake`...

@4a6f656c any opinions on this?

update.sh
busterb: sure

busterb:
> I'd be inclined to leave `LIBRESSL_VERSION_NUMBER` as `0x4040000fL` the same way as in the actual release and consider modifying the `_TEXT` with a patch (this might require moving this awk line to after the patching if that's possible). This way things that really feel they must support and special case an rc can inspect the `_TEXT`. But that might break `FindLibreSSL.cmake`...

I think I agree with you leave NUMBER as-is, add RC to _TEXT so folks can tell they're on an rc somehow in the built artifacts.

botovq (approved):
Let's try this. @4a6f656c wants the first rc release to be rc0.

busterb:
```
bcook@bobcat ~/projects/libressl/portable/libressl-4.3.1rc0
 % LIBRESSL_RC=0 ./dist.sh > /dev/null
bcook@bobcat ~/projects/libressl/portable
 % tar zxf libressl-4.3.1rc0.tar.gz
bcook@bobcat ~/projects/libressl/portable
 % cd libressl-4.3.1rc0
bcook@bobcat ~/projects/libressl/portable/libressl-4.3.1rc0
 % ./configure > /dev/null
bcook@bobcat ~/projects/libressl/portable/libressl-4.3.1rc0
 % make -j8 > /dev/null
ld: warning: netcat.c:475(netcat.o:(main)): warning: mktemp() possibly used unsafely; consider using mkstemp()
bcook@bobcat ~/projects/libressl/portable/libressl-4.3.1rc0
 % ./apps/openssl/openssl version
WARNING: can't open config file: /usr/local/etc/ssl/openssl.cnf
LibreSSL 4.3.1rc0
```

PR #1395, opened 2026-09-20
split pipe2 from socketpair detection

busterb:
Support macOS 27's new pipe2 that can build but not work if you have SDK 27 installed on 26. This verifies it actually works before linking it. Split socketpair from pipe2, since they can now be independently available.


busterb:
cmake and some other stuff coming in a sec - not expecting this to totally work on the first try.

botovq:
@sage-mode-hunter would you mind taking a look at this?

busterb:
I could be convinced that we should just let it fail, and that it should be up to the developer to select the right SDK level when building, but this does prevent an out of the box footgun that didn't exist in previous releases. And we do need the split when actually targeting macOS 27 natively.

busterb:
Are we good on this one? This is needed for a clean build on a couple of my machines now. Definitely needed if you update to Golden Gate

tests/CMakeLists.txt
```diff
@@ -803,16 +803,49 @@ else()
...
+# signertest and tlstest are the only consumers of compat/pipe2.c and
+# compat/socketpair.c; check for both here since they're needed by both
+# targets below.
+# Emscripten does not support the socketpair syscall.
 if(NOT (CMAKE_SYSTEM_NAME MATCHES "WindowsStore" OR EMSCRIPTEN))
-	set(SIGNERTEST_SRC signertest.c)
-	check_symbol_exists(pipe2 "fcntl.h;unistd.h" HAVE_PIPE2)
+	# SDK headers can declare pipe2() as available even if the underlying libc
+	# only has a stub, e.g. macOS 27 SDK running on macOS 26.
+	# SDK) ahead of the running OS actually implementing it.
```
nak3: This line is a little difficult for me to understand. Could you clarify what it means?

nak3 (approved):
I left one minor comment but looks good to me.
I also tested this with the xcode-27 GitHub Actions runner. The configure logs showed:

xcode-27:

```
2026-09-23T23:48:25.7291440Z checking for working pipe2... yes
```

macos-26:

```
2026-09-23T23:48:36.2843640Z checking for working pipe2... no
```

This confirms that CI exercises both paths, although I was not able to test the specific combination of the SDK 27 on macOS 26.

botovq:
No objection from my side, although I haven't reviewed it carefully and won't have time for that in the coming days.

tests/CMakeLists.txt
busterb: When I tested this initially, I tested on a macOS 26 machine, that had auto-updated to the macOS 27 SDK. Everything builds, but it crashes by default because macOS 26 doesn't implement pipe2(), and building in this way links to a broken libc stub for a non-existent syscall.

nak3: Thank you, that explanation makes sense. I was referring specifically to the wording of this line:

```
# SDK) ahead of the running OS actually implementing it.
```

It looks like an incomplete sentence or a leftover fragment. Would it make sense to keep it, even though the preceding two lines already explain the issue?

PR #1396, opened 2026-09-20
simplify asm detection with cmake on x64 Windows

### LINKED ISSUE #1379: CMakeLists.txt doesn't reliably detect x64 when using MSVC
`crypto/CMakeLists.txt` has the following:

	if(MSVC)
		if((NOT "${CMAKE_GENERATOR}" MATCHES "Win64") AND
		   (NOT "${CMAKE_GENERATOR_PLATFORM}" STREQUAL "x64"))
			add_definitions(-DOPENSSL_NO_ASM)
		endif()
	elseif(WIN32 AND NOT MINGW)
		add_definitions(-DOPENSSL_NO_ASM)
	endif()

I often use the "NMake Makefiles" generator with CMake and when doing so, the logic quoted above will identify the "NMake Makefiles" generator as NOT an x64 target and subsequently add the `-DOPENSSL_NO_ASM` definition which means that all people using the "NMake Makefiles" generator will get builds that have asm disabled which is probably not desirable.

So I think there should be some more robust detection if the target platform is x64 than just relying on the generator's name.

busterb:
We already determine in a more build-system-agnostic way when to enable asm, so just use those definitions rather than reimplementing here. I can't remember why we did it this way originally, maybe it was to workaround an old CMake issue?

This should fix https://github.com/libressl/portable/issues/1379


botovq (approved):
Makes sense to me.

joshuasing (approved):

PR #1397, opened 2026-09-20
readme: document versioning scheme

joshuasing:
Add a Versioning section to the `README.md` file that explains the versioning scheme, with the addition of release candidates (related to https://github.com/libressl/portable/pull/1393).


busterb:
Do we want to mention that historically we haven't done this until 4.4.0, or just document as though we had?

busterb (approved):

joshuasing:
> Do we want to mention that historically we haven't done this until 4.4.0, or just document as though we had?

I think it could be useful to mention this. @botovq what are your thoughts on this?

botovq (approved):

PR #1400, opened 2026-09-21
Update endian.h for compatibility with llvm-mingw 23

oold:
llvm-mingw is shipping with an endian.h header since LLVM 23. The current preprocessor check breaks the build with that version because neither is endian.h included nor does the block that defines the byte swap functions inline become active since it's guarded by `defined(_WIN32) && !defined(HAVE_ENDIAN_H)`, so there's no way the byte swap functions are ever defined when building for Windows and an endian.h header is detected.


botovq (approved):
Thanks!

PR #1404, opened 2026-09-22
return -1 from poll when WaitForMultipleObjects fails

sage-mode-hunter:
poll() reports a failed WaitForMultipleObjects as a timeout instead of an error:
- the WAIT_FAILED branch sets errno = ENOMEM then returns 0, but poll(2) uses 0 for "timed out, nothing ready", so the caller never inspects errno and the wait failure reads as a benign timeout
- every other error exit in the same function returns -1 (NULL pfds, the two FD_SETSIZE caps, and the select() path via wsa_select_errno)
- an invalid or closed non-socket fd yields INVALID_HANDLE_VALUE from _get_osfhandle, so WaitForMultipleObjects returns WAIT_FAILED and an openssl(1) s_client/s_server poll loop spins on the masked failure

Return -1 so the error surfaces; errno is already set.


botovq (approved):

joshuasing (approved):

joshuasing:
Thanks!

